PR #2, opened 2024-02-09
Write tests for BoundaryTrie functionality

emdoyle:


modguard/core/boundary.py
```diff
@@ -24,21 +24,26 @@ class BoundaryTrie:
     def __iter__(self):
         return boundary_trie_iterator(self)
 
+    @staticmethod
+    def _split_mod_path(path: str) -> list[str]:
```
caelean, line 28:
```suggestion
    def _split_mod_path(path: str, sep='.') -> list[str]:
```
if we ever want to do something like `_split_mod_path(..., "/")

tests/test_boundary_trie.py
```diff
@@ -0,0 +1,160 @@
...
+    trie = BoundaryTrie()
+    assert trie.get("") is None
+
+
+def test_get_actual_path(boundary_trie):
+    assert boundary_trie.get("domain_one") is not None
+
+
+def test_get_actual_empty_path(boundary_trie):
+    assert boundary_trie.get("") is not None
```
caelean, line 83: could be nice to have a stricter assertion than `is not None` ie check against the expected returned value

caelean (approved):

PR #4, opened 2024-02-10
More test cases/data

caelean:
- hit more branches
- updated test coverage cases
- ruff check passing without ignore rule
- more usages
- more test cases  Please enter the commit message for your changes. Lines starting


tests/example/domain_four/subsystem.py
```diff
@@ -1,7 +1,8 @@
...
+import modguard
+import math
 
-Boundary()
+modguard.Boundary()
 
 
 def private_subsystem_call():
-    ...
+    print(math.random())
```
emdoyle, line 8: I think this moved to `random.random`

PR #6, opened 2024-02-12
Handle missing dynamic imports

emdoyle:
If `mark_as_public` fails to find the specified member_name in the file_path, we previously raised an error.
Now we will add a call to `public()` with the member_name used as a string literal argument. This is also now recognized by the public member visitor.

### DIFF
--- a/modguard/init.py
+++ b/modguard/init.py
@@ -81,8 +81,11 @@ def init_project(root: str, exclude_paths: Optional[list[str]] = None):
                 # This import is fine, no need to mark anything as public
                 continue
 
-            file_path, member_name = fs.module_to_file_path(import_mod_path)
+            member_name = ""
             try:
+                file_path, member_name = fs.module_to_file_path(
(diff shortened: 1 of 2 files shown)


modguard/init.py
```diff
@@ -103,4 +106,8 @@ def init_project(root: str, exclude_paths: Optional[list[str]] = None):
             elif write_op.operation == WriteOperation.PUBLIC:
                 mark_as_public(write_op.location, write_op.member_name)
         except errors.ModguardError:
-            print(f"Error marking {write_op.operation} in {write_op.location}")
+            print(
+                f"Error marking {write_op.operation.value}"
+                f"{'({member})'.format(member=write_op.member_name) if write_op.member_name else ''}"
```
caelean, line 111: never seen a format call inside an f string lol might be worth doing this formatting outside of the print() call

caelean (approved):

PR #8, opened 2024-02-13
basic show command

caelean:
Basic show command, including an option to write it to yaml

```bash
modguard show [-w] .
```

### DIFF
--- a/modguard/cli.py
+++ b/modguard/cli.py
@@ -95,7 +95,7 @@ def parse_show_arguments(args: list[str]) -> argparse.Namespace:
     parser.add_argument(
         "path",
         type=str,
-        help="The path of the Python project in which boundaries should be initialized.",
+        help="The path of the Python project in which to show boundaries and public members.",
     )
     parser.add_argument(
         "-e",
@@ -184,7 +184,7 @@ def modguard_init(args: argparse.Namespace):
 def main() -> None:
     if len(sys.argv) > 1 and sys.argv[1] == "init":
         modguard_init(parse_init_arguments(sys.argv[2:]))
-    if len(sys.argv) > 1 and sys.argv[1] == "show":
+    elif len(sys.argv) > 1 and sys.argv[1] == "show":
         modguard_show(parse_show_arguments(sys.argv[2:]))
     else:
         modguard(parse_base_arguments(sys.argv[1:]))
--- a/modguard/cli.py
+++ b/modguard/cli.py
@@ -4,6 +4,8 @@ import sys
 
 from modguard.check import check, ErrorInfo
 from modguard.init import init_project
+from modguard.show import show
+from modguard.parsing.boundary import build_boundary_trie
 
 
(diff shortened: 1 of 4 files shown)


modguard/show.py
```diff
@@ -0,0 +1,38 @@
+import yaml
+from modguard.core.boundary import BoundaryTrie
+from typing import Any, Dict
+
+# This type hint only works on more recent versions
+# result_dict: TypeAlias = dict[str, str | bool | 'result_dict']
+
+def boundary_trie_to_dict(boundary_trie: BoundaryTrie) -> Dict[str, Any]:
```
caelean, line 8: Typing w/ Any here because the recursively nested types weren't passing earlier python versions

modguard/cli.py
```diff
@@ -142,6 +184,8 @@ def modguard_init(args: argparse.Namespace):
 def main() -> None:
     if len(sys.argv) > 1 and sys.argv[1] == "init":
         modguard_init(parse_init_arguments(sys.argv[2:]))
+    if len(sys.argv) > 1 and sys.argv[1] == "show":
```
emdoyle, line 187: nit: prefer elif

modguard/show.py
```diff
@@ -0,0 +1,42 @@
...
+        path = node.full_path
+        if path == "":
+            path = "root"
+        sections = path.split(".")
+        pointer: Dict[str, Any] = result
+        for section in sections:
+            if section not in pointer:
+                pointer[section] = dict()
+            pointer = pointer[section]
+        pointer["is_boundary"] = True
```
emdoyle, line 20: nit: 'pointer' is kinda confusing, given usage maybe 'current' is better?

modguard/show.py
```diff
@@ -0,0 +1,42 @@
...
+
+        for member in node.public_members.keys():
+            pointer: Dict[str, Any] = result
+            sections = member.split(".")
```
emdoyle, line 24: Since the root boundary `__init__.py` may be marked `public()` and would then have a public member with an empty path, we probably want to explicitly handle that case and assign `member` to "root" like you do in the previous loop.

modguard/show.py
```diff
@@ -0,0 +1,42 @@
+import yaml
+from modguard.core.boundary import BoundaryTrie
+from typing import Any, Dict
+
+# This type hint only works on more recent versions
+# result_dict: TypeAlias = dict[str, str | bool | 'result_dict']
+
+def boundary_trie_to_dict(boundary_trie: BoundaryTrie) -> Dict[str, Any]:
```
emdoyle, line 8: We may run into issues in the case where part of the file structure uses the name `is_public` or `is_boundary`, which would be weird but not that unusual.

modguard/show.py
emdoyle: We could also ignore the root boundary altogether, not sure it adds anything to the output

PR #25, opened 2024-02-23
Revert single-level restriction on module-level public() call

emdoyle:
Reverts back to earlier behavior of `public()`, implicitly marking all descendants of the calling module as public on the nearest boundary.

Updates tests & docs to reflect this behavior.


docs/api.md
```diff
@@ -70,44 +70,51 @@ modguard.public(x, allowlist=["project.core.domain"])
...
-
-@modguard.public(allowlist=["project.core.domain"])
-def my_pub_function():
-    ...
-```
-
-### Entire Module
-When `public` is used without a `path` argument, it signifies that the entire containing module is public. This means that any top-level member of the module or the module itself can be imported externally (subject to `allowlist`).
+### Module-level
+When `public` is used without a `path` argument (the most common case), it signifies that the containing module and its descendants are public. This means that any descendant of the module or the module itself can be imported externally (subject to `allowlist`). Note that adding a `Boundary` in a descendant module would prevent that module from being treated as public by default.
```
caelean, line 74:
```suggestion
When `public` is used without a `path` argument (the most common case), it signifies that the containing module and its descendants are public. This means that any descendant of the module or the module itself can be imported externally (subject to `allowlist`). Note that adding a `Boundary` in a descendant module will prevent that module from being treated as public by default.
```

caelean (approved):

PR #32, opened 2024-04-25
Core Rewrite

emdoyle:
Moving from in-line config (@public, Boundary()) to explicit external config (modguard.yml, module.yml).

Updating check logic and core data structures to operate on 'tags' and dependency constraints.


modguard/check.py
```diff
@@ -1,94 +1,94 @@
...
-        else None
-    )
-    import_mod_is_public_and_allowed = (
-        import_mod_public_member_definition is not None
-        and (
-            import_mod_public_member_definition.allowlist is None
-            or check_allowlist(
-                import_mod_public_member_definition.allowlist, file_mod_path
+        if "*" in dependency_tags:
+            continue
```
caelean, line 73: We prob want to support `re.match` here too

caelean (approved):

Issue #75, opened 2024-05-23
python 3.8 support?


brianmedigate:
Does this project support python 3.8? From the `pyproject.toml` it appears that it does, but the codebase uses python-3.9-style type annotations which subscript the builtin `dict` type.

```
$ tach -h
Traceback (most recent call last):
  File "/home/brianm/repos/medigator/venv/bin/tach", line 5, in <module>
    from tach.cli import main
  File "/home/brianm/repos/medigator/venv/lib/python3.8/site-packages/tach/cli.py", line 8, in <module>
    from tach.check import check, BoundaryError
  File "/home/brianm/repos/medigator/venv/lib/python3.8/site-packages/tach/check.py", line 5, in <module>
    from tach import filesystem as fs
  File "/home/brianm/repos/medigator/venv/lib/python3.8/site-packages/tach/filesystem/__init__.py", line 1, in <module>
    from tach.filesystem.service import (
  File "/home/brianm/repos/medigator/venv/lib/python3.8/site-packages/tach/filesystem/service.py", line 46, in <module>
    def _get_file_cache() -> dict[str, FileInfo]:
TypeError: 'type' object is not subscriptable
```

```
$ python -V
Python 3.8.19
```

```
$ pip show tach
Name: tach
Version: 0.2.2
...
```

emdoyle:
Thanks for raising this!

@max-muoto fixed this :muscle: and it's now released as `0.2.3`!

PR #76, opened 2024-05-23
Python 3.8 Support

max-muoto:
Add Python 3.8 support by deferring annotations, which should allow new-style annotations to be utilized in older versions of Python, including the pipe operator for optionals.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
 
     steps:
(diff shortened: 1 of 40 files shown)


max-muoto:
@caelean or @emdoyle Not able to run the entire test suite locally due a missing `.example` directory, but I think this should fix https://github.com/gauge-sh/tach/issues/75 if one of you can approve the CI jobs.

emdoyle:
:pray: Thanks for doing this!

> @caelean or @emdoyle Not able to run the entire test suite locally due a missing `.example` directory, but I think this should fix #75 if one of you can approve the CI jobs.

This is actually because our pytest configuration is wonky, you need to be in the `tests` directory to run them. It was likely looking for the `example` directory which is at `tests/example` from the root. Regardless, CI runs the tests so this is looking good!

emdoyle (approved):
:snake:

PR #77, opened 2024-05-23
logging

caelean:
Adds a custom remote logger handler; updates faq


tach/cache/setup.py
```diff
@@ -0,0 +1,43 @@
...
+    return Path(current_dir)
+
+
+def resolve_dot_tach() -> Path:
+    project_dir = resolve_project_dir()
+
+    def _create(path: str, is_file: bool = False, file_content: str = "") -> None:
+        if not os.path.exists(path):
+            if is_file:
+                with open(path, "w") as f:
```
max-muoto, line 22: Using pathlib, things like this can be a little simpler:

```python
path = Path("df")
path.write_text(file_content.strip())
```


It will basically open up a context manager for you behind the scenes.

tach/logging/api.py
```diff
@@ -0,0 +1,36 @@
...
+
+def log_request(url: str, data: dict) -> None:
+    headers = {
+        "Content-Type": "application/json",
+        "apikey": PUBLIC_ANON_CLIENT_KEY,
+        "authorization": f"Bearer {PUBLIC_ANON_CLIENT_KEY}",
+    }
+    try:
+        response = requests.post(
+            f"{LOGGING_URL}/{url}", data=json.dumps(data), headers=headers
```
max-muoto, line 23: Probably good to set a timeout here.

max-muoto: https://github.com/psf/requests/issues/3070

tach/cache/setup.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
+
+import os
+import uuid
+from pathlib import Path
+
+
+def resolve_project_dir() -> Path:
+    current_dir = os.getcwd()
```
max-muoto, line 9: This should work:

```python
from pathlib import Path

current_dir = Path.cwd()
tach_path = current_dir / "tach.yml"
```

`pathlib` basically wraps everything you'd want to use in `os`.

caelean:
@max-muoto thanks for the review 🙏 I like pathlib's interface; will try and rip and replace more `os` stuff moving forward.

max-muoto:
> @max-muoto thanks for the review 🙏 I like pathlib's interface; will try and rip and replace more `os` stuff moving forward.

There's a nice Ruff rule-set for this: https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth, which is helpful in terms of finding the replacements!

tach/cache/access.py
```diff
@@ -0,0 +1,15 @@
...
+
+import os
+import uuid
+
+from tach.cache.setup import resolve_dot_tach, resolve_project_dir
+
+
+def get_uid() -> uuid.UUID:
+    project_root = resolve_project_dir()
+    if not os.path.exists(project_root / ".tach" / "tach.info"):
```
max-muoto, line 11: I think this works: `if not (project_root / ".tach" / "tach.info").exists():`

caelean: nice, updated/tested

tach/cache/__init__.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import annotations
```
emdoyle, line 1: Does this need to be here? I haven't used future statements before but maybe it's better to add it only to files that use annotations, or if it applies once globally at import time, I think we can put it in `tach/__init__.py`

tach/cache/setup.py
```diff
@@ -0,0 +1,40 @@
...
+    project_dir = resolve_project_dir()
+
+    def _create(path: Path, is_file: bool = False, file_content: str = "") -> None:
+        if not path.exists():
+            if is_file:
+                path.write_text(file_content.strip())
+            else:
+                path.mkdir()
+
+    # Create .bridge
```
emdoyle, line 25:
```suggestion
    # Create .tach
```

tach/logging/logger.py
```diff
@@ -0,0 +1,57 @@
...
+logger = logging.getLogger("tach")
+logger.setLevel(logging.INFO)
+remote_handler = RemoteLoggingHandler()
+
+# Check if logging is enabled
+disable_logging = False
+try:
+    project_config = parse_project_config()
+    enable_logging = project_config.disable_logging
+except SystemExit:
```
emdoyle, line 52: We should really pull the `sys.exit` out of `validate_project_config_path`

tach/logging/logger.py
```diff
@@ -0,0 +1,57 @@
...
+    }
+    log_uid(uid, is_ci)
+    log_record(log_data)
+
+
+class RemoteLoggingHandler(logging.Handler):
+    def emit(self, record: logging.LogRecord) -> None:
+        log_entry = self.format(record)
+        thread = threading.Thread(target=send_log_entry, args=(record, log_entry))
+        thread.start()
```
emdoyle, line 40: I think this should be a daemon thread so that it can't cause the process to hang.

https://docs.python.org/3/library/threading.html#threading.Thread.daemon

tach/cache/setup.py
```diff
@@ -0,0 +1,40 @@
...
+import uuid
+from pathlib import Path
+
+
+def resolve_project_dir() -> Path:
+    current_dir = Path.cwd()
+    tach_path = current_dir / "tach.yml"
+    if not tach_path.exists():
+        raise FileNotFoundError()
+    return Path(current_dir)
```
emdoyle, line 12: Should this use `find_project_config_root`? Also I think we should handle errors in the thread target (`send_log_entry` probably) since the error traces will still hit stdout otherwise.

tach/cache/package.yml
```diff
@@ -0,0 +1 @@
+tags: ['tach.cache']
```
emdoyle, line 1: This would also affect tach.yml
```suggestion
tags: ['scope:cache']
strict: true
```

tach/logging/logger.py
caelean: 💯

tach/cache/__init__.py
caelean: It's a requirement as of #76 

```
[tool.ruff.lint.isort]
required-imports = ["from __future__ import annotations"]
```

tach/logging/logger.py
```diff
@@ -0,0 +1,94 @@
...
+
+logger = logging.getLogger("tach")
+logger.setLevel(logging.INFO)
+remote_handler = RemoteLoggingHandler()
+
+# Check if logging is enabled
+disable_logging = False
+try:
+    project_config = parse_project_config()
+    enable_logging = project_config.disable_logging
```
emdoyle, line 88:
```suggestion
    disable_logging = project_config.disable_logging
```

emdoyle (approved):
One comment to fix the opt-out, otherwise looks good!

tach/cache/__init__.py
max-muoto: It's just easier to have it everyone from my experience, some big codebases like https://github.com/apache/airflow do this.

tach/logging/logger.py
caelean: Ah great catch

PR #79, opened 2024-05-24
Enable TCH Rule-Set

max-muoto:
Enable TCH rule-set so that to avoid unnecessary imports at module init time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,10 +41,21 @@ Issues = "https://github.com/gauge-sh/tach/issues"
 
 [tool.ruff]
 target-version = "py38"
-lint.extend-select = ["I"]
+lint.extend-select = ["I", "TCH"]
 
 [tool.ruff.lint.isort]
 required-imports = ["from __future__ import annotations"]
+
+[tool.ruff.lint.flake8-type-checking]
+runtime-evaluated-base-classes = [
+    "pydantic.BaseModel",
+    "pydantic.RootModel",
+]
+runtime-evaluated-decorators = [
+    "pydantic.dataclasses.dataclass",
+    "dataclasses.dataclass",
+]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -41,11 +41,22 @@ Issues = "https://github.com/gauge-sh/tach/issues"
...
 target-version = "py38"
-lint.extend-select = ["I"]
+lint.extend-select = ["I", "TCH"]
 
 [tool.ruff.lint.isort]
 required-imports = ["from __future__ import annotations"]
 
+[tool.ruff.lint.flake8-type-checking]
+runtime-evaluated-base-classes = [
+    "pydantic.BaseModel",
```
max-muoto, line 51: Basically ignore fields being used on dataclass-like objects.

caelean: @emdoyle would be nice to get all the `@dataclass` implementations over to pydantic eventually too

caelean (approved):

pyproject.toml
max-muoto: I'm a huge fan of msgspec, it sorted of designed for statically typed codebases (like tach) where they know validation isn't necessary in internal code-paths where you have type-checking enabled, but it lets you validate external data (whether from config, JSON, etc).

max-muoto: Initialization type is even faster than dataclasses from the testing I've done. https://jcristharif.com/msgspec/benchmarks.html

emdoyle: > @emdoyle would be nice to get all the `@dataclass` implementations over to pydantic eventually too

I think pydantic is useful for config since it validates user-editable values, but dataclasses + pyright should be lower overhead and give us the same value for data that we control end-to-end internally. Never heard of msgspec though, might be a good candidate to replace dataclasses :eyes:

PR #81, opened 2024-05-24
Reconfigure the repo to build a Python package with a Rust extension library

emdoyle:
:crab: :crab: :crab:

This PR is the first step toward extracting the core logic of `tach` to Rust.

The motivation is primarily performance, we want to be able to run `check` frequently, even against large repos, in order to support use cases like an IDE extension.

This PR introduces `maturin` as our build backend, which sets up bindings to Rust with `pyo3`.

Python source code is now located in `python/` while Rust source code is located in `src/`
There is now a `Cargo.toml` alongside `pyproject.toml` in the project root.

NOTE: This PR bumps the version to `0.2.6`, and also fixes a few bugs that I encountered while testing - primarily around the `--root` argument to check


.github/workflows/ci.yml
```diff
@@ -35,20 +35,21 @@ jobs:
...
+        ruff check
+        ruff format --check
     - name: Test with pytest and report coverage
       run: |
         source .venv/bin/activate
-        cd tests
+        cd python/tests
         coverage run --branch --source=../ -m pytest
         coverage report
-        cd ..
```
caelean, line 46: Oh nice, does gh actions do some cleanup between steps? This was actually always a pain in circleci

.github/workflows/ci.yml
```diff
@@ -35,20 +35,21 @@ jobs:
...
     - name: Check types with pyright
       run: |
         source .venv/bin/activate
-        pyright .
+        cd python
+        pyright
     - name: Check tach
       run: |
-        pip install .
-        tach check --exclude tests,build
```
caelean, line 54: Thoughts on adding the basic cargo ci stuff? `clippy`/`check`/`fmt` etc

.github/workflows/publish.yml
```diff
@@ -11,21 +13,120 @@ permissions:
...
+            target: armv7
+          - runner: ubuntu-latest
+            target: s390x
+          - runner: ubuntu-latest
+            target: ppc64le
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
```
caelean, line 37: Looks like this is auto-generated, I assume this won't be an issue when used with other python versions that we support?

caelean (approved):

.github/workflows/ci.yml
emdoyle: Yeah I looked into this, seems like each step starts a fresh shell from the project root, while the filesystem is shared across all steps.

.github/workflows/publish.yml
emdoyle: My understanding was that since we delegate the build to `maturin`, any Python version which it supports should function basically the same way. I'm not sure if there is an affect on the built wheels though, in my head it should basically be Python source + compiled Rust but I will look into it further.

emdoyle: Looked into this more, the ABI can change across Python versions, so we actually do we need to build wheels for every combo of OS/arch/Python that we want to support, but it looks like our GHA runners already have CPython 3.8-12 and a few versions of PyPy. Maturin auto-detects all of these and builds wheels for them (due to `--find-interpreter`)

PR #84, opened 2024-05-28
Basic Makefile with Python Support

max-muoto:
Basic makefile with support for running linting checks and installing Python dependencies.


emdoyle (approved):
Thanks! :pray:

PR #87, opened 2024-05-29
Only require eval type backport < 3.10

max-muoto:
This library is only needed for Python syntax features that weren't present before 3.10, so we can safely ignore it for later versions of Python, reducing dependencies for some users.


caelean (approved):
Thanks Max!

Issue #96, opened 2024-06-01
Dependencies need to be relaxed


looztra:
The way the dependencies of Tach are defined is very strict (only using `==`)

```toml
dependencies = [
    "pyyaml==6.0.1",
    "pydantic==2.7.1",
    "rich==13.7.1",
    "prompt-toolkit==3.0.43",
    "stdlib-list==0.10.0; python_version < '3.10'",
    "eval-type-backport==0.2.0; python_version < '3.10'"
]
```

This means that folks that want to install tach have to have these exact dependencies or it won't install.
As tach will most of the time not be installed in its own venv, this is problematic.

Could you be open to a PR that would relax the dependencies?

looztra:
By the way, using a tool like poetry, rye or pdm could help in order to manage the runtime + dev dependencies...

PR #97, opened 2024-06-01
chore: loosen dependency requirements

max-muoto:
For most libraries simply require they're not a major version away, since `stdlib-list` and `eval-type-backport` really shouldn't change their interface from version to version just require any version above what we have now. We could probably lighten the requirements even further, once we implement improved testing from major version to version, but this at least improves things in the meantime.

Fixes: https://github.com/gauge-sh/tach/issues/96


emdoyle (approved):
Thanks for this!

PR #109, opened 2024-06-12
Warn users on older versions

caelean:
- Cache latest version in `.tach`
- Check current version against latest when cli is run and notify user to upgrade as needed
- Add `tach --version` and `tach.__version__`
- Make `cli`, `mod`, `clean`, and `sync` strict modules

### DIFF
--- a/python/tach/cache/access.py
+++ b/python/tach/cache/access.py
@@ -13,8 +13,8 @@ def get_uid() -> uuid.UUID | None:
         return
     if not (project_path / ".tach" / "tach.info").exists():
         resolve_dot_tach()
-    with open(project_path / ".tach" / "tach.info") as f:
-        uid = uuid.UUID(f.read().strip())
+    contents = (project_path / ".tach" / "tach.info").read_text().strip()
+    uid = uuid.UUID(contents)
     return uid
 
 
@@ -24,8 +24,7 @@ def get_latest_version() -> str | None:
         return
     if not (project_path / ".tach" / ".latest-version").exists():
         return
-    with open(project_path / ".tach" / ".latest-version") as f:
-        version = f.read().strip()
+    version = (project_path / ".tach" / "latest-version").read_text().strip()
     return version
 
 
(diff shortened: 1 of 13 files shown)


python/tach/cache/access.py
```diff
@@ -1,19 +1,47 @@
...
-    if project_root is None:
+    project_path = get_project_path()
+    if project_path is None:
         return
-    project_path = Path(project_root)
     if not (project_path / ".tach" / "tach.info").exists():
         resolve_dot_tach()
-    with open(project_path / ".tach" / "tach.info") as f:
-        uid = uuid.UUID(f.read().strip())
+    contents = (project_path / ".tach" / "tach.info").read_text().strip()
```
emdoyle, line 16: nit: would be nice to put the `project_path / ...` into a variable

python/tach/cache/access.py
```diff
@@ -1,19 +1,47 @@
...
+    try:
+        # Sending a GET request to the URL
+        with request.urlopen(url, timeout=1) as response:
+            if response.status == 200:
+                data = response.read().decode()
+                json_data = json.loads(data)
+                latest_version = json_data["info"]["version"]
+            else:
+                return
+    except (error.URLError, KeyError):
```
emdoyle, line 45: might also be good to catch errors related to JSON parsing or utf-8 decoding as well to avoid printing them to stdout when the tool exits (although they should be unlikely)

python/tests/test_cache.py
```diff
@@ -0,0 +1,118 @@
...
+
+@patch("tach.cache.setup.find_project_config_root")
+def test_get_project_path_no_root(mock_find_project_config_root):
+    mock_find_project_config_root.return_value = None
+    result = get_project_path()
+    assert result is None
+
+
+@patch("tach.cache.setup.get_project_path")
+def test_resolve_dot_tach(mock_get_project_path, tmp_path):
```
emdoyle, line 28: TIL about `tmp_path`, seems useful!

emdoyle (approved):
Nice! Works well when testing locally.

PR #110, opened 2024-06-13
Add `tach report` command

emdoyle:
This PR adds the `tach report` command.

This command can be run from anywhere within a tach project, and will generate a 'dependency report' for the given filepath.

This is separate from module configuration and in fact doesn't care whether the filepath corresponds to any module. It is simply intended to quickly see the dependencies and usages of a given file or directory.

This is useful if you want to understand how 'self-contained' a particular module is, or how much work it might be to isolate and extract a module.

Example output from `tach report tach/cache/`:

```
[Dependency Report for tach/cache/]
The report below shows all instances of imports which cross the boundary of 'tach/cache/'
-------------------------------
[Dependencies of 'tach/cache/']
tach/cache/setup.py[L6]: Import 'tach.__version__'
tach/cache/setup.py[L7]: Import 'tach.filesystem.find_project_config_root'
-------------------------------
[Usages of 'tach/cache/']
tach/cli.py[L11]: Import 'tach.cache'
tach/logging/logger.py[L11]: Import 'tach.cache'
```


README.md
```diff
@@ -95,6 +95,14 @@ tach clean
 This will wipe all the configuration generated and enforced by Tach.
 
 
+If you just want to see the external dependencies and external usage of a given file or module in your project, you can use `tach report`.
+
+```bash
+tach report my_package
```
caelean, line 101:
```suggestion
tach report my_package/
```
Just to clearly differentiate from the other use case

caelean: Wonder if we'd want to support list of items here similar to `--exclude` in the future as well

README.md
```diff
@@ -95,6 +95,14 @@ tach clean
 This will wipe all the configuration generated and enforced by Tach.
 
 
+If you just want to see the external dependencies and external usage of a given file or module in your project, you can use `tach report`.
```
caelean, line 98: Imo `external` is pretty confusing here - when I think external, I think third party dependencies or usages outside of my project, not outside of the boundary. Here's the first result from google - 
image
I think simply just saying "dependencies and usages outside of the [boundary/module]" makes sense for here, the argparse docstring, and the output of running the command.

caelean (approved):
Looks good, just one main comment on naming convention here. Could see a few more nice to haves for this in the feature, like supporting lists/globs/generating a report for everything/writing the report to a file/etc none of which make sense now but just worth writing out for the future! Great work

PR #115, opened 2024-06-14
Show warnings when file parsing fails during project import gathering

emdoyle:
This PR changes the behavior of Tach in the presence of syntax errors.

Previously, the entire process would exit with an error message indicating a syntax error.

With this PR, Tach will instead report a warning to stdout when the syntax error is encountered during `check` or `sync`, or will collect warnings and add them to the output of `tach report`.

### DIFF
--- a/python/tach/check.py
+++ b/python/tach/check.py
@@ -6,7 +6,6 @@ from typing import TYPE_CHECKING
 
 from tach import errors
 from tach import filesystem as fs
-from tach.colors import BCOLORS
 from tach.extension import get_project_imports, set_excluded_paths
 from tach.parsing import build_module_tree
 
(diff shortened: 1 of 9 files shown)


caelean (approved):

python/tach/check.py
```diff
@@ -149,11 +150,22 @@ def check(
...
+                )
+            except SyntaxError:
+                print(
+                    f"{BCOLORS.WARNING}Skipping file '{file_path}' due to a syntax error.{BCOLORS.ENDC}"
+                )
+                continue
+            except OSError:
+                print(
+                    f"{BCOLORS.WARNING}Skipping file '{file_path}' due to a file system error.{BCOLORS.ENDC}"
+                )
```
caelean, line 167:
```suggestion
                    f"{BCOLORS.WARNING}Skipping '{file_path}' due to a syntax error.{BCOLORS.ENDC}"
                )
                continue
            except OSError:
                print(
                    f"{BCOLORS.WARNING}Skipping '{file_path}' due to a file system error.{BCOLORS.ENDC}"
                )
```
small nit, `file` is prob self explanatory for a file path

PR #121, opened 2024-06-20
[Refactor] Add explicit concept of 'source_root' to Tach

emdoyle:
image

This PR introduces the concept of a Python 'Source Root' as a first-class concept in Tach.
You can configure your source root using `tach mod` (shown above) or manually by adding it to `tach.yml`.

This also means that projects should almost always put `tach.yml` in their git root, and configure their `source_root` as necessary, instead of needing to colocate `tach.yml` with their Python source code.

Along with these changes, this PR moves the vast majority of our usage of `str` for file paths to the more appropriate `pathlib.Path`. This cascaded into changes in quite a few files, and is significantly more robust than the original implementation.

More importantly, this PR eliminates a class of issues that users could face related to their current working directory. Now, Tach will automatically search for a `tach.yml` file in ancestors of the cwd to determine the project root, and will use absolute paths throughout the core logic. This removes the need for the `--root` option passed to `tach check` (`source_root` makes this obsolete), and allows running Tach's commands from anywhere in your project.

If it seems like your source root is not set correctly, Tach will show a warning with a link to the source root documentation:
image

This PR also updates and fixes our tests so that they can be run from the root of the repo, instead of being dependent on the current working directory (no more `cd python/tests`).

Docs have been updated in this PR, but a version bump is left to be done separately.


README.md
```diff
@@ -75,11 +77,13 @@ You will see:
 
 You can validate that Tach is working by either commenting out an item in a `depends_on` key in `tach.yml`, or by adding an import between modules that didn't previously import from each other. 
 
-Give both a try and run `tach check` again. This will generate an error:
+Give both a try and run `tach check` again. This will generate an error similar to this one:
 ```bash
-❌ path/file.py[LNO]: Cannot import 'path.other'. Tags ['scope:other'] cannot depend on ['scope:file']. 
+❌ tach/check.py[L8]: Cannot import 'tach.filesystem'. Tag 'tach' cannot depend on 'tach.filesystem'.
```
caelean, line 82: Do you think using `tach` as an example within `tach` is a bit too confusing? I do think the arbitrary one isn't great either.

One push I got was to update our readme with a motivating example - something for us to noodle more on.

docs/configuration.md
```diff
@@ -32,12 +32,50 @@ exclude:
...
 - tests/
+source_root: backend
 exact: true
 disable_logging: false
 ignore_type_checking_imports: true
 ```
 
-## modules
+## Source Root
+The `source_root` key is critically important for Tach to understand the imports within your project.
```
caelean, line 42:
```suggestion
The `source_root` key is necessary for Tach to understand the imports within your project.
```

docs/configuration.md
```diff
@@ -32,12 +32,50 @@ exclude:
...
+```python
+# In backend/module1.py
+
+import module3
+from module2.service import MyService
+```
+
+Notice that these import paths (`module3`, `module2.service.MyService`) are rooted in the `backend/` folder, NOT the project root.
+
+To indicate this structure to Tach, you could set `source_root: backend` in your `tach.yml`, or use [`tach mod`](usage.md#tach-mod) and interactively mark the `backend` folder as the source root.
```
caelean, line 74:
```suggestion
To indicate this structure to Tach, set `source_root: backend` in your `tach.yml`, or use [`tach mod`](usage.md#tach-mod) and interactively mark the `backend` folder as the source root.
```

docs/usage.md
```diff
@@ -1,5 +1,35 @@
...
+You can navigate with the arrow keys, mark individual modules with `Enter`, and mark all siblings
+as modules with `Ctrl + a`.
+
+You can also mark your Python [source root](configuration.md#source-root) by pressing `s`.
+This allows Tach to understand the module paths used in your project, and to correctly identify first-party imports.
+
+When you are ready to save your modules, use `Ctrl + s` to save and exit. Otherwise, to exit without saving you can use `Ctrl + c`.
+
+Any time you make changes with `tach mod`, it is recommended to run [`tach sync`](usage.md#tach-sync)
+to automatically configure dependency rules.
```
caelean, line 30: It might be nice to have a little table in the format

Command: action

I have an outstanding task to do another pass on the docs, so def nonblocking

python/tach/check.py
```diff
@@ -131,90 +131,106 @@ class ProjectModuleValidationResult:
...
+                    error_info=check_error,
                 )
+            )
 
-        return CheckResult(errors=boundary_errors, warnings=warnings)
-    finally:
-        fs.chdir(cwd)
+    if not found_at_least_one_project_import:
+        warnings.append(
+            "WARNING: No first-party imports were found. You may need to use 'tach mod' to update your Python source root. Docs: https://gauge-sh.github.io/tach/configuration#source-root"
```
caelean, line 231: nice 👏

python/tach/core/config.py
```diff
@@ -57,10 +58,15 @@ class ProjectConfig(Config):
     """
 
     modules: list[ModuleConfig] = Field(default_factory=list)
-    exclude: list[str] | None = Field(default_factory=lambda: ["tests", "docs"])
+    exclude: list[str] | None = Field(default_factory=lambda: ["tests", "docs", "venv"])
```
caelean, line 61: might be nice to pull these defaults into a constant we can reference elsewhere

python/tach/filesystem/install.py
```diff
@@ -1,15 +1,16 @@
...
 
+if TYPE_CHECKING:
+    from pathlib import Path
 
-def install_pre_commit(path: str = ".", project_root: str = "") -> tuple[bool, str]:
-    root_path = Path(path)
 
-    git_hooks_dir = root_path / ".git/hooks"
+def install_pre_commit(project_root: Path) -> tuple[bool, str]:
+    git_hooks_dir = project_root / ".git/hooks"
```
caelean, line 13:
```suggestion
    git_hooks_dir = project_root / ".git" / "hooks"
```
for windows compat

python/tach/interactive/modules.py
```diff
@@ -427,6 +470,16 @@ def _(event: KeyPressEvent):
                 # Root should not be explicitly selected
                 return
             self.selected_node.is_module = not self.selected_node.is_module
+            if self.selected_node.is_module and self.selected_node.is_source_root:
+                # If we are marking the currently selected source root as a module,
+                # we should reset the source root to the root path.
+                # The source root cannot be a module.
+                self.file_tree.set_source_root(Path("."))
```
caelean, line 477: Makes sense, I wonder if setting source root as parent vs root path could make more sense, but would have to play with it a bit more. I guess we could kind of figure it out by looking at your imports but that's way more work; this is def a sane default.

caelean (approved):
Looks great - love the pathlib updates + expanded exclude, the python source root makes a ton of sense. Only had a chance to skim the code so haven't pulled it down and played with it yet, but will give it a try soon!

README.md
emdoyle: I think for now it's OK but agree we should have a more concrete, relatable example here.

PR #122, opened 2024-06-20
Add color coding to tach report

caelean:
- **fmt**
- **new colors module**
- **register mod**
- **implement colors**
- **clippy/fmt**
- **make command to develop rust**
- **correctly wrap colors, endc**

### DIFF
--- a/src/reports.rs
+++ b/src/reports.rs
@@ -1,6 +1,8 @@
 use std::cmp::Ordering;
 use std::fmt::{self, Debug};
 
+use crate::colors::*;
+
 use crate::filesystem::{
     adjust_path_from_cwd_to_root, file_to_module_path, walk_pyfiles, FileSystemError,
 };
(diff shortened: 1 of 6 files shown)


src/imports.rs
```diff
@@ -321,7 +327,7 @@ pub fn get_project_imports(
     ignore_type_checking_imports: bool,
 ) -> Result<ProjectImports> {
     let file_path = PathBuf::from(&file_path);
-    let absolute_source_root = PathBuf::from(&project_root).join(&source_root);
+    let absolute_source_root = PathBuf::from(&project_root).join(source_root);
```
emdoyle, line 330: This is a surprising change, since I'd typically always use a reference when possible. Did a formatting tool suggest this?

src/imports.rs
```diff
@@ -40,6 +40,12 @@ pub struct ProjectImport {
     pub line_no: u32,
 }
 
+pub struct Dependency {
+    pub file_path: PathBuf,
+    pub absolute_path: PathBuf,
+    pub import: ProjectImport,
+}
```
emdoyle, line 47: nit: I'd expect this to stay as a private struct in reports.rs since all its construction and usage is in that file

src/reports.rs
```diff
@@ -124,24 +124,29 @@ impl DependencyReport {
             {subtitle}\n\
             -------------------------------\n\
             [{deps_title}]\n\
-            {deps}\n\
+            {cyan}{deps}{end_color}\n\
             -------------------------------\n\
             [{usages_title}]\n\
-            {usages}",
+            {green}{usages}{end_color}",
```
emdoyle, line 130: nit: The output seems hard to read for me with these colors, maybe we could distinguish the file link from the import with something like yellow (or plain white) for all the `Import: ...` sections?

image

emdoyle (approved):
:crab: 

Working for me! Left some comments, nonblocking.

src/imports.rs
caelean: Yeah, this is all clippy

src/imports.rs
caelean: Ah whoops, moved it out as I was originally using it in `cli.rs` but then changed that - will move back

Issue #125, opened 2024-06-21
Gracefully exit on KeyboardInterrupt


emdoyle:
Right now, if you send a keyboard interrupt (`Ctrl + c`) during Tach's execution, you'll get an ugly stack trace.

Instead, we should catch a KeyboardInterrupt at a high level and exit gracefully (with a nonzero exit code).

nikhil5642:
Hey @emdoyle, the ugly stack trace is due to 2 reason's
1. Impot's taking time to load
2. Tach acutall task execution is not catching keyboard exception.

We can fix the 2nd cause by just adding a simple try catch on main function inside cli.py but inoder to fix this issue completely we will have to go for one hackey solution.

We can fix the first issue by creating a new file let's say(start.py), and call this inside pyproject.toml. So instead of calling tach.cli:main we will call tach.start:start.

This is how this start.py will look like
`from __future__ import annotations

import sys
if __name__ == "__main__":
    try:
        from tach.cli import main
        main()
    except KeyboardInterrupt:
        print("KeyboardInterrupt")
        sys.exit(1)
  `

PR #129, opened 2024-06-23
Fixes Gracefully exit on `KeyboardInterrupt`

legendkartik45:
This PR fixes https://github.com/gauge-sh/tach/issues/125

Description:
This PR enhances the user experience by catching KeyboardInterrupt exceptions during Tach's execution. Currently, interrupting the script with Ctrl + C results in an ugly stack trace. Instead, this PR introduces a try-except block in the main script to catch KeyboardInterrupt at a high level. Upon receiving a KeyboardInterrupt, the script will exit gracefully with a nonzero exit code, ensuring a cleaner termination.

Changes:

`Earlier`:
image

`Now`
image

@emdoyle Any feedbacks on the above PR from your side would be appreciated

### DIFF
--- a/python/tach/__main__.py
+++ b/python/tach/__main__.py
@@ -1,6 +1,10 @@
 from __future__ import annotations
-
+import sys
 from tach.cli import main
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except KeyboardInterrupt:
+        print("\nExiting gracefully...")
+        sys.exit(1)


nikhil5642:
@legendkartik45 I tried this solution, It won't solve this issue. The main cause of this issue is the import's, that are taking time to load. I order to fix this issue you will have to move all the import's inside cli.py. 

Here in __main__.py, one hacky way to solve it is to move `from tach.cli import main` inside try catch, but that will only solve issue for __main__py.

Another hackey but more efficient way to solve this is to create another file(say start.py) and change parth to start.py in pyproject.toml.

If you check pyproject.toml we are directly consuming main funciton inside tach.cli. So when when you build project with 'maturin develop' it will auto generate this file. So your solution will clearly won't work.

#!/Users/nikhagrawal/opt/anaconda3/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from tach.cli import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())

emdoyle:
@legendkartik45 Thanks for taking a stab at this issue! This would work in some cases but we should tweak the approach a bit.
@nikhil5642 You're right that we need this behavior around `cli.main` as well, and we need to catch a keyboard interrupt during import loading.

The reason `__main__.py` exists is so that we can execute Tach as a module (`python -m tach`) which is required in some environments, like the VSCode extension. Generally, using Tach on the CLI will use the entrypoint from `pyproject.toml`, which points directly at the `main` function in `cli.py`.

To get the behavior we want in both cases, I like the idea of a `start.py`, which could look like:
```python
import sys

def start():
    try:
        from tach.cli import main
        
        main()
    except KeyboardInterrupt:
        print("Exiting...")
        sys.exit(1)

if __name__ == "__main__":
    start()
```

And then we could update `pyproject.toml` to reference `tach.start:start`, and update the contents of `__main__.py` to:
```python
from tach.start import start

start()
```

I think we can also remove the `if __name__ == "__main__"` block from `cli.py`, since we don't expect that file to be executed as a script on its own.

emdoyle:
Appreciate your contribution here @legendkartik45 , but closing this PR in favor of #131 which was just merged.

Issue #127, opened 2024-06-21
Move '<root>' module to the bottom/end of modules in tach.yml


emdoyle:
The implicit root boundary appears in the `modules` list in `tach.yml` with the path:  `<root>`

Right now, when this module exists, it is always at the beginning of the `modules` list (sorted lexicographically). This is not ideal, because as a reader (especially a new user), the meaning of `<root>` is hard to parse before first seeing the rest of the modules. This is because a reasonable definition of the root boundary is the 'leftover' code - any code which is not covered by another module.

Given this, it is easier to read the explicit modules first, and then see the 'leftovers' at the end of the file. Tach should explicitly sort this module last when writing the project config to yml.

buddhiraz:
I have just pushed one to fix this arrangement, plz revieew . I havent added description in the PR yet.

dhananjaypai08:
Can you give me an example use case of what the tach.yml might look before and after this enhancement

emdoyle:
@dhananjaypai08 sure! here is an example. I have removed a lot of extra content so that it is easier to read, but the only real difference is that right now, the `<root>` module (if it exists) appears first, and it should instead appear last.

```
modules:
- path: <root>  # root module is first
  depends_on:
  - tach.filesystem
- path: tach.check
  depends_on:
  - <root>
  - tach.errors
  - tach.filesystem
  - tach.parsing
  strict: true
- path: tach.cli
  depends_on:
  - <root>
  - tach.check
  strict: true
...
```

```
modules:
- path: tach.check
  depends_on:
  - <root>
  - tach.errors
  - tach.filesystem
  - tach.parsing
  strict: true
- path: tach.cli
  depends_on:
  - <root>
  - tach.check
  strict: true
- path: <root>  # root module should be last
  depends_on:
  - tach.filesystem
...
```

dhananjaypai08:
This is a easy fix by just finding the selected root folder in the config.modules list and then popping it and appending to the same list in the config::dump_project_config_to_yaml. But I don't know why in my case in the tach.yaml there is no `-path: <root module>` anywhere. I tried it using in different cases(via my own build as well as through the actual released library) and every time the `-path: <root>` the root is not added in the tach.yaml file

emdoyle:
[...]

PR #130, opened 2024-06-23
Trying to fix the str of tach.yml

buddhiraz:
Fix for [issue 127](https://github.com/gauge-sh/tach/issues/127) 

Have added a function in sync.py : reorder_modules
- which keeps the ordering in check (as requested)
```Python
def reorder_modules(project_config: ProjectConfig) -> None:
    modules = project_config.modules
    root_module = next((module for module in modules if module.name == "<root>"), None)
    if root_module:
        modules.remove(root_module)
        modules.append(root_module)
    project_config.modules = modules
```
- also added a test_sync.py to test the modified sync.py .


tach.yml
```diff
@@ -1,107 +0,0 @@
-modules:
```
emdoyle, line 1: This file should not be deleted, it is how we configure `tach` for our own repo.

run_sync.py
```diff
@@ -0,0 +1,31 @@
+
```
emdoyle, line 1: This file should be removed, I'm not sure what it's for. Running `tach sync` is done through the `cli.py` file based on command-line arguments.

python/tests/test_sync.py
```diff
@@ -0,0 +1,49 @@
+import unittest
```
emdoyle, line 1: Tests for this repository should be written using the `pytest` framework, you can check some of the sibling files for examples. Given the content of this file, it would probably be better to add these tests for `reorder_modules` into `test_parsing` -- also because `reorder_modules` should be moved into the `parsing` module instead of `sync`.

python/tach/sync.py
```diff
@@ -90,9 +85,9 @@ def sync_project(
             project_config=project_config,
             exclude_paths=exclude_paths,
         )
-
+    reorder_modules(project_config)
     tach_yml_content = dump_project_config_to_yaml(project_config)
```
emdoyle, line 89: Instead of adding this logic as a separate function and calling it in `tach sync`, it would be better to include this behavior in `dump_project_config_to_yaml`, so that we get consistent sorting whenever we write the config file to disk. You'll see there is already a normal sort of the modules happening there, so you could use a custom comparator function to special-case the root module, or simply move the root module to the end after the sort is done.

emdoyle (changes requested):
Thanks for looking into this! I think the approach should be modified, and I left some comments about how the repo is structured. I'll collect some of this information and prepare a `CONTRIBUTING.md` and guide in the docs soon.

python/tach/sync.py
buddhiraz: Yes I think that I can work on.

PR #131, opened 2024-06-23
Keyboard ugly stack trace handled

nikhil5642:
Fix for issue: https://github.com/gauge-sh/tach/issues/125

the ugly stack trace is due to 2 reason's
1. Impot's taking time to load
2. Tach acutall task execution is not catching keyboard exception.

I am crating a PR that will fix both the issues by moving both import and actual tach execution inside a Try Catch. This is done by creating another file start.py and moving the entry from main class of cli.py to start class of start.py

From now onward's this is how execution will exit.
    
Screenshot 2024-06-24 at 3 30 04 AM

This is how our start.py will look like:
```python
from __future__ import annotations

import sys

def start():
    try:
        from tach.cli import main
        main()
    except KeyboardInterrupt:
        print("Exiting...")
        sys.exit(1)
        
if __name__ == "__main__":
    start()

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,4 +79,4 @@ module-name = "tach.extension"
 features = ["pyo3/extension-module"]
 
 [project.scripts]
-tach = "tach.cli:main"
+tach = "tach.start:start"
--- a/python/tach/__main__.py
+++ b/python/tach/__main__.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from tach.cli import main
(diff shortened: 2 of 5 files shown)


emdoyle:
This looks good! It just looks like it needs to be formatted by `ruff` to pass the CI check, and the `tach.yml` configuration should be updated.

emdoyle (approved):
Thanks for this! It will go out in the next release (0.6.2) later today.

PR #138, opened 2024-06-26
Move '<root>' module to the bottom/end of modules in tach.yml

### LINKED ISSUE #127: Move '<root>' module to the bottom/end of modules in tach.yml

dhananjaypai08:
The implicit boundary occurs in very certain use cases, therefore haven't extensively tested this enhancement
Fixes #127 
let me know what you think

### DIFF
--- a/python/tach/parsing/config.py
+++ b/python/tach/parsing/config.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import yaml
 
 from tach import filesystem as fs
+from tach.constants import ROOT_MODULE_SENTINEL_TAG
 from tach.core import ProjectConfig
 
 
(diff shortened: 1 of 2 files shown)


python/tach/parsing/config.py
```diff
@@ -14,7 +14,10 @@ def dump_project_config_to_yaml(config: ProjectConfig) -> str:
     # Instead, should provide custom yaml.Dumper & yaml.Representer or just write our own
     # Sort only constraints and dependencies alphabetically for now
     config.modules.sort(key=lambda mod: mod.path)
```
emdoyle, line 16: This approach worries me due to the concurrent modification (modifying `config.modules` while iterating over it). I think it would be more natural and more efficient to adjust the sort we are doing on the line before, something like:

```
config.modules.sort(key=lambda mod: (mod.path == ROOT_MODULE_SENTINEL_TAG, mod.path))
```
which would use a tuple as a key to sort the root module to the end (since Python sorts False before True as if they were ints).

dhananjaypai08: Gotcha, better approach than concurrent modification of file list while traversing.  Currently used `(bool, module.path)` within config.modules.sort where the only time bool value is True is when its equal to root and all the root paths move to the bottom in a sorted way

emdoyle (approved):
Looks great! Just needs ruff to pass, should be just `ruff check --fix`.

PR #139, opened 2024-06-27
tach test

emdoyle:
image

## tach test
This PR introduces `tach test` - a test runner that uses module boundaries to perform [test impact analysis](https://martinfowler.com/articles/rise-test-impact-analysis.html) and only run the tests impacted by your changes. It also uses an on-disk cache to restore test results when possible.

### Implementation
- Test impact analysis logic using `git` to determine changed files, and then propagating those changes through the dependencies defined in `tach.yml`
- A `pytest` plugin which uses affected module paths and imports from each collected test file to skip unaffected tests (with clear reporting)
- Several command line options for customizing `tach test` behavior, including  the ability to forward arbitrary arguments to the underlying `pytest` command (ex: `tach test -- -v` will forward `-v` to `pytest`)
- Add functions to Rust extension to generate cache keys, get cached results and store cached results
- Context manager in cli.py to capture stdout and stderr output during execution, and preserve their interleaved order for accurate replay on cache hits
- Additional `cache` config key available in `tach.yml` which offers additional glob patterns for file dependencies (aside from Python files within the source root), environment variable dependencies, and a `backend` option which is currently unused (but will allow selecting a remote cache in the future)

### TODO
- [x] Tests
- [x] Documentation


emdoyle:
Found a couple things during testing that I'll address before merging:
- by default shouldn't cache if pytest collection fails (should only cache if tests actually run)
- pytest collection fails if tests exist outside of source root -- need to actually handle get_project_imports in this case (ignore relative imports if running from outside the project?)

PR #148, opened 2024-07-05
prune the default behaviour and --add for adding missing dependency

### LINKED ISSUE #147: Make --prune the default behavior of tach sync
Right now, the default behavior of `sync` is additive, meaning that it will not remove any modules or configured dependencies. Instead it will add dependencies to existing modules to fix any boundary errors detected.

When using the `--prune` option, `sync` ignores the currently configured dependencies, and instead sets the dependencies for each module to the minimal set matching the actual imports found in the codebase. This behavior is more consistent with the idea of "syncing" your dependencies, and should be the default behavior when calling `tach sync` with no CLI arguments.

The current additive behavior should be enabled by passing a renamed flag: `--add`
That is, `tach sync` will prune dependencies to the minimal set, while `tach sync --add` will add any missing dependencies.

NOTE: This issue interacts with #145 , and may require coordination if both are worked on simultaneously

dhananjaypai08:
Fixes #147 
Note: Not been extensively tested, may require changes and corresponding doc changes


python/tach/cli.py
```diff
@@ -123,9 +123,9 @@ def print_unused_dependencies(
         + constraint_messages
     )
     print(
-        f"{BCOLORS.WARNING}\nRemove the unused dependencies from tach.yml, "
-        f"or consider running 'tach sync --prune' to update module configuration and "
-        f"eliminate all unused dependencies.\n{BCOLORS.ENDC}"
+        f"{BCOLORS.WARNING}\nAdd the unused dependencies from tach.yml, "
+        f"or consider running 'tach sync --add' to update module configuration and "
+        f"add all unused dependencies.\n{BCOLORS.ENDC}"
     )
```
emdoyle, line 129:
```suggestion
        f"{BCOLORS.WARNING}\nRemove the unused dependencies from tach.yml, "
        f"or consider running 'tach sync' to update module configuration and "
        f"remove all unused dependencies.\n{BCOLORS.ENDC}"
    )
```

emdoyle: In this case, we are showing an error to the user because they have unused dependencies in their `tach.yml`, so our advice should still be to remove those dependencies.

The part that changes is that since `--prune` is now the default, we can just tell them to run `tach sync`

python/tach/cli.py
```diff
@@ -191,9 +191,9 @@ def build_parser() -> argparse.ArgumentParser:
         description="Sync constraints with actual dependencies in your project.",
     )
     sync_parser.add_argument(
-        "--prune",
+        "--add",
         action="store_true",
-        help="Prune all existing constraints and re-sync dependencies.",
+        help="Add all existing constraints and re-sync dependencies.",
```
emdoyle, line 196:
```suggestion
        help="Add any missing dependencies, but do not remove unused dependencies.",
```

emdoyle (commented):
Thanks for tackling this, functionally it looks good!

I left some suggestions for the messages shown to the user. Also, I think there should be documentation changes in docs/usage.md and anywhere else we give examples of using `tach sync --prune`.

emdoyle (approved):
Looks good, thanks! :rocket:

PR #151, opened 2024-07-08
feat: Preserve language server directive and formatting in tach sync

### LINKED ISSUE #150: Maintain formatting and language server directive in tach.yml after sync
Right now, the `tach.yml` in our repo is formatted with 2 space indents at each level of nesting, and there is a language server comment directive at the top of the file.

After running any variant of `tach sync` however, the language server directive is eliminated, and the formatting reverts to be minimally indented.

It would be much nicer if `tach sync` could preserve formatting, but at a minimum it should preserve the language server directive.

tiluckdave:
- [x] Retained the language server comment directive at the top of tach.yml
- [x] Implemented a custom YAML dumper `TachYamlDumper` to maintain 2-space indentation for consistent formatting

Fixes #150

### DIFF
--- a/python/tach/parsing/config.py
+++ b/python/tach/parsing/config.py
@@ -8,7 +8,7 @@ from tach import filesystem as fs
 from tach.constants import ROOT_MODULE_SENTINEL_TAG
 from tach.core import ProjectConfig
 
-class CustomDumper(yaml.Dumper):
+class TachYamlDumper(yaml.Dumper):
     def increase_indent(self, flow=False, indentless=False):
         return super().increase_indent(flow, False)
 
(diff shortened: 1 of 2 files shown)


python/tach/parsing/config.py
```diff
@@ -24,7 +27,14 @@ def dump_project_config_to_yaml(config: ProjectConfig) -> str:
     # show excluded paths.
     config.exclude = list(set(config.exclude)) if config.exclude else []
     config.exclude.sort()
-    return yaml.dump(config.model_dump(exclude_unset=True), sort_keys=False)
+    language_server_directive = "# yaml-language-server: $schema=docs/assets/tach-yml-schema.json\n"
```
emdoyle, line 30: In this case, the path `docs/assets/tach-yml-schema.json` will only work within our repo. External users typically use our hosted version: `https://gauge-sh.github.io/tach/assets/tach-yml-schema.json`

Could you move that URL into a constant and reference the constant here?

tiluckdave: Yes sure!
I had more more doubt is there a right place for creating the TachYamlDumper class. I found it odd being in the same file.

emdoyle (approved):
LGTM

python/tach/parsing/config.py
emdoyle: Great thanks! And for the placement, I don't see an issue, since the `parsing` module is responsible for the YAML serialization/deserialization and this is the only place the dumper would be used.

PR #153, opened 2024-07-08
Bump pyo3

### LINKED ISSUE #152: Bump pyo3 to 0.22 to fix build issue on 3.13
```
...
         Compiling pyo3-macros-backend v0.20.3
         Compiling pyo3-ffi v0.20.3
         Compiling pyo3 v0.20.3
         Compiling globset v0.4.14
      error: failed to run custom build command for `pyo3-ffi v0.20.3`

      Caused by:
        process didn't exit successfully: `C:\Users\ךינשגכהד\AppData\Local\Temp\pip-req-build-8ntdecot\target\release\build\pyo3-ffi-a96f43da0592a7f0\build-script-build` (exit code: 1)
        --- stdout
        cargo:rerun-if-env-changed=PYO3_CROSS
        cargo:rerun-if-env-changed=PYO3_CROSS_LIB_DIR
        cargo:rerun-if-env-changed=PYO3_CROSS_PYTHON_VERSION
        cargo:rerun-if-env-changed=PYO3_CROSS_PYTHON_IMPLEMENTATION
        cargo:rerun-if-env-changed=PYO3_PRINT_CONFIG
        cargo:rerun-if-env-changed=PYO3_USE_ABI3_FORWARD_COMPATIBILITY

        --- stderr
        error: the configured Python interpreter version (3.13) is newer than PyO3's maximum supported version (3.12)
        = help: please check if an updated version of PyO3 is available. Current version: 0.20.3
        = help: set PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 to suppress this check and build anyway using the stable ABI
      warning: build failed, waiting for other jobs to finish...
      💥 maturin failed
        Caused by: Failed to build a native library through cargo
        Caused by: Cargo build finished with "exit code: 101": `"cargo" "rustc" "--features" "pyo3/extension-module" "--message-format" "json-render-diagnostics" "--manifest-path" "C:\\Users\\ךינשגכהד\\AppData\\Local\\Temp\\pip-req-build-8ntdecot\\Cargo.toml" "--release" "--lib" "--crate-type" "cdylib"`
      Error: command ['maturin', 'pep517', 'build-wheel', '-i', 'C:\\Users\\ךינשגכהד\\scoop\\apps\\python-alpha\\current\\python.exe', '--compatibility', 'off'] returned non-zero exit status 1
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for tach
Failed to build tach
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (tach)
```

clin1234:
Fixes #152


emdoyle (commented):
Changes look good, but it looks like PyO3 changed its API significantly in `0.21`. Could you also make the relevant changes in `lib.rs`? It seems like they should be relatively straightforward substitutions and shouldn't leak out of that file.

https://pyo3.rs/v0.22.1/

emdoyle:
@clin1234 Your changes look good, but if I'm reading the failure output correctly, it looks like `pydantic-core==2.18.2` doesn't distribute a pre-built wheel for 3.13, and it needs an incompatible version of pyo3 (0.21.1). This requirement must be coming from `pydantic` which we have as a floating version.

I would recommend upping our pydantic version requirement when python_version is 3.13, since the latest version of pydantic and pydantic-core seem to be using a later version of pyo3.

clin1234:
I do not know where that specific version of pydantic-core is pulled from: latest release is https://github.com/pydantic/pydantic-core/releases

clin1234:
Perhaps you should setup dependabot support so that manual updates to dependencies won't be necessary...

emdoyle:
> Perhaps you should setup dependabot support so that manual updates to dependencies won't be necessary...

We may in the future, but the change you are trying to make here is purely to support Python 3.13, which is a beta release and not a target version that Tach supports today, so I don't think Dependabot would have changed anything here.

Your latest commit updates the dev requirements, but the issue here is the package requirement in `pyproject.toml`:
`"pydantic~=2.0"`

This dependency is purposefully loose so that we can avoid unnecessary conflicts in users' Python projects, since `pydantic` is a very common dependency. Upgrading it to force a late enough version of `pydantic-core` to support the beta release of Python 3.13 is not a good tradeoff in general, so the only reasonable change I could see here is a specific version upgrade when the detected python version is 3.13 or higher. You can see an example of using the python version to specify dependencies also in `pyproject.toml`:
```
    "stdlib-list>=0.10.0; python_version < '3.10'",
    "eval-type-backport>=0.2.0; python_version < '3.10'"
```

If you can specify a workable version range for Python 3.13 and avoid impacting other users, that would work.

clin1234:
> > Perhaps you should setup dependabot support so that manual updates to dependencies won't be necessary...
> 
> We may in the future, but the change you are trying to make here is purely to support Python 3.13, which is a beta release and not a target version that Tach supports today, so I don't think Dependabot would have changed anything here.
> 
> Your latest commit updates the dev requirements, but the issue here is the package requirement in `pyproject.toml`: `"pydantic~=2.0"`
> 
> This dependency is purposefully loose so that we can avoid unnecessary conflicts in users' Python projects, since `pydantic` is a very common dependency. Upgrading it to force a late enough version of `pydantic-core` to support the beta release of Python 3.13 is not a good tradeoff in general, so the only reasonable change I could see here is a specific version upgrade when the detected python version is 3.13 or higher. You can see an example of using the python version to specify dependencies also in `pyproject.toml`:
> 
> ```
>     "stdlib-list>=0.10.0; python_version < '3.10'",
>     "eval-type-backport>=0.2.0; python_version < '3.10'"
> ```
> 
> If you can specify a workable version range for Python 3.13 and avoid impacting other users, that would work.

I haven't modified pyproject.toml, where most users would find a suitable version of `pydantic`; I only bumped the version of `pydantic` within the dev requirements, in order to fix a specific dep for building on 3.13 through CI. This shouldn't impact the workflow of a casual user.

emdoyle:
> I haven't modified pyproject.toml, where most users would find a suitable version of `pydantic`; I only bumped the version of `pydantic` within the dev requirements, in order to fix a specific dep for building on 3.13 through CI. This shouldn't impact the workflow of a casual user.

I see what you're saying - all Rust dependencies should be build-time/dev dependencies assuming we pre-build wheels for all supported environments. In that case, there may be another change required to the `publish` workflow. It uses the `--find-interpreter` option to determine which Python interpreters it will build wheels for. This depends on the interpreters installed in our GitHub Actions runner, which I believe are shown here: https://github.com/gauge-sh/tach/actions/runs/9848383661/job/27190215219#step:4:394

This means we will not build wheels for 3.13, so if you are planning to use this version you'll need to manage the Rust toolchain on your local machine as well - or you'll need to modify the `publish.yml` to setup Python 3.13 (and also it looks like we are missing 3.7)

I'm happy to merge this as-is and deal with the wheel change separately, or if you want to make the GHA change and use this branch that's fine too!

emdoyle (approved):

clin1234:
Good enough. I'll make the publish.yml change in a different branch

PR #169, opened 2024-07-10
fails `tach check` when circular dependencies

### LINKED ISSUE #146: Add setting to fail 'tach check' if circular dependencies exist
Some projects would like to ensure that their project's dependencies remain unidirectional, or in other words, that there are no cycles in their dependency graph.

A simple way to begin supporting this is to add a setting like `forbid_circular_dependencies` which can be set in `tach.yml` (`ProjectConfig` within the Tach codebase). When this setting is true, `tach check` should fail with a specific error message when a circular dependency is detected.

This should likely be done as a "pre-flight check" which would validate the `modules` list, and could be delegated to the `ModuleTree` internally. It does not need to apply any new logic to actual detected imports.

tiluckdave:
Checks for circular dependencois if `forbid_circular_dependencies` is set to true in tach.yml file.

Fixes #146


tiluckdave:
How do I create tests for this?

emdoyle:
> How do I create tests for this?

Tests would be awesome. We use the `python/tests` folder to hold our tests, and we use `pytest` (or `tach test`) to run them.

There should be a few good examples there, like this one: https://github.com/gauge-sh/tach/blob/3f91c1da08862473a06118ebf4e33b95f7fd5ee8/python/tests/test_imports.py

which set up project files in a temporary directory and then check that Tach's logic produces the expected results.

Building a few test cases with and without cycles would be fantastic!

tiluckdave:
@emdoyle Can you please review!

python/tach/parsing/modules.py
```diff
@@ -22,12 +22,51 @@ def find_duplicate_modules(modules: list[ModuleConfig]) -> list[str]:
...
+    return []
+
+
+def find_modules_with_circular_dependencies(modules: list[ModuleConfig]) -> list[str]:
+    modules_with_cycles: set[str] = set()
+    for module in modules:
+        cycle = find_cycle(module, set(), [], modules)
+        if cycle:
+            modules_with_cycles.update(cycle)
+    return list(modules_with_cycles)
```
emdoyle, line 53: If I understand correctly, this collapses all discovered cycles into a single set. This might make it hard for a user to understand where the cycle(s) exist in their dependencies.

Maybe instead we could have this function return a list of all discovered cycles, and the caller can raise a custom exception which holds this data?

That way, we can catch the custom exception somewhere like `cli.py` and print a nice colorful error message with all the detected cycles.

python/tach/parsing/modules.py
```diff
@@ -22,12 +22,51 @@ def find_duplicate_modules(modules: list[ModuleConfig]) -> list[str]:
     return duplicate_module_paths
 
 
-def build_module_tree(source_root: Path, modules: list[ModuleConfig]) -> ModuleTree:
+def find_cycle(
```
emdoyle, line 25: Could you add a few test cases for this function?

emdoyle (commented):
This is looking pretty good! Just had a couple asks to improve the UX and add tests before merging.

python/tach/parsing/modules.py
tiluckdave: I created a custom error in error module.
Is it okay if I make the error module import colors?

python/tach/parsing/modules.py
tiluckdave: I believe the tests should go in `test_parsing.py`

tiluckdave:
Hey @emdoyle can you please review

python/tests/test_parsing.py
```diff
@@ -54,3 +55,31 @@ def test_invalid_project_config(example_dir):
...
+        visited: set[str] = set()
+        path: list[str] = list()
+        find_cycle(module, visited, path, modules, all_cycles)
+    assert all_cycles == {
+        ("domain_one", "domain_three", "domain_one"),
+        ("domain_three", "domain_one", "domain_three"),
+        ("domain_three", "domain_one", "domain_two", "domain_three"),
+        ("domain_two", "domain_three", "domain_one", "domain_two"),
+        ("domain_one", "domain_two", "domain_three", "domain_one"),
+    }
```
emdoyle, line 85: Thanks for the tests!

This is not the output I would expect from finding cycles in a graph though, and would probably be confusing to a user. It seems like finding 'unique' cycles in general might best be done with `networkx`. This dependency was added just recently in #182. Here is some example code that shows the approach we should adopt:

```
import networkx as nx

def canonical_form(cycle):
    """Rotate cycle to start with the smallest element."""
    n = len(cycle)
    min_index = cycle.index(min(cycle))
    return cycle[min_index:] + cycle[:min_index]

def find_unique_cycles(...):
    # Build a networkx.DiGraph
    # use the code in show.py:generate_module_graph_dot_file as a model for this
    
    # Find all simple cycles in the graph
    all_cycles = list(nx.simple_cycles(G))
    
    # Normalize cycles to their canonical form
    canonical_cycles = {tuple(canonical_form(cycle)) for cycle in all_cycles}
    
    # Convert tuples back to lists
    unique_cycles = [list(cycle) for cycle in canonical_cycles]
    
    return unique_cycles
```

tach.yml
```diff
@@ -47,7 +47,8 @@ modules:
       - tach.constants
     strict: true
   - path: tach.errors
-    depends_on: []
+    depends_on:
+      - tach.colors
```
emdoyle, line 51: See other comment, we should avoid adding this dependency

emdoyle (commented):
Thanks for the changes and especially for the tests! Given that output, I think we should just use networkx.

tiluckdave:
@emdoyle I am done with the requested changes.



~ Sorry for the delay. I was unwell for the past week.

tach.yml
```diff
@@ -1,4 +1,4 @@
-# yaml-language-server: $schema=https://raw.githubusercontent.com/gauge-sh/tach/v0.6.9/public/tach-yml-schema.json
+# yaml-language-server: $schema=https://raw.githubusercontent.com/gauge-sh/tach/v0.8.2/public/tach-yml-schema.json
```
caelean, line 1: The public schema also needs to be updated with `forbid_circular_dependencies`! You can find it at `public/tach-yml-schema.json`

python/tach/parsing/modules.py
```diff
@@ -22,12 +25,46 @@ def find_duplicate_modules(modules: list[ModuleConfig]) -> list[str]:
...
+) -> list[list[str]]:
+    graph = nx.DiGraph()  # type: ignore
+    # Add nodes
+    for module in modules:
+        graph.add_node(module.path)  # type: ignore
+
+    # Add dependency edges
+    for module in modules:
+        for dependency in module.depends_on:
+            graph.add_edge(module.path, dependency)  # type: ignore
```
caelean, line 45: small nit - it looks like building the networkx Digraph is something we're doing in multiple places now - it might make sense to have it be a property or function of `ProjectConfig`, so we're not duplicating code to build the graph

caelean (changes requested):
Hey @tiluckdave, sorry for the delays in reviewing. This is awesome and looking super close to being ready to merge - appreciate the work here!

python/tach/parsing/modules.py
tiluckdave: Makes sense @caelean . Just one doubt here the module list we are using is validated using `fs.validate_project_modules` in `check.py:154`

If we make it a property of ProjectConfig it will use original modules not the validated ones.

```py
@property
def dependency_graph(self) -> nx.DiGraph:
    graph = nx.DiGraph()  # type: ignore
    # Add nodes
    for module in self.modules:
        graph.add_node(module.path)  # type: ignore

    # Add dependency edges
    for module in self.modules:
        for dependency in module.depends_on:
            graph.add_edge(module.path, dependency)  # type: ignore
    
    return graph
```

I created this property!

caelean:
@tiluckdave just made a couple final tweaks in #206 - merging this there! Thank you so much for the contribution 🙏

PR #189, opened 2024-07-17
Fix deps make function, install pip and maturin

letmerecall:
## Why this PR?
`make deps` for fresh venv where pip and maturin are not installed.

## What's updated?
Added a step to install `pip` & `maturin` in deps make function.


emdoyle (approved):
Great catch, thanks for the fix!

PR #207, opened 2024-07-28
Support `deprecated` dependencies

### LINKED ISSUE #184: Add deprecated modifier to file
I have the following use case:

I want to slowly introduce tach and thus slowly remove depends_on from some modules. If I add a comment in the tach.yml it gets overwritten by sync so desirable behaviour for me would be:

Either:
- Do not delete comments
- Have a deprecation annotation 
  - tach could support checking violations with all deprecated dependencies removed and without them being removed)
  - automatically remove deprecation annotations on sync if the dependency is not present anymore

caelean:
Add support for specifying a dependency is `deprecated`. This involves a breaking change to the `tach.yml` spec:
From:
```yaml
modules:
  - path: domain_one
    depends_on:
      - domain_two
      - domain_three
```
to
```yaml
modules:
  - path: domain_one
    depends_on:
      - path: domain_two
        deprecated: true
      - path: domain_two
 ``` 
(if marking `domain_two` as a deprecated dependency)

Here's what it looks like in action!
image

This is necessary to support per-dependency configuration, which we'll likely expand in the future. Included is a migration handler which will auto-update every project config to the new format when it's run. 
Note that for now we'll still lose comments from the file when this happens.    

`deprecated` spits out a warning, but still returns the exit code 0. I debated adding a flag to fail on deprecations, but for now you can achieve the same thing by just removing the dependency from `depends_on`.

This is also a breaking change for `tach show --web` - I'll set up changes on the web server to go out alongside the release of this feature. Fixes #184!

Bonus - expanded the test suite a little bit and added caching for deps! Takes it from about 3 min -> 1 min

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,4 +65,6 @@ jobs:
       run: cargo fmt --all --check
     - name: Check Rust
       run: cargo check
+    - name: Compile Maturin
+      run: maturin develop --release -E dev
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,10 +50,11 @@ jobs:
       run: |
         ruff check
         ruff format --check
-    - name: Test with pytest and report coverage
+    - name: Test with tach/pytest and report coverage
       run: |
         coverage run --branch --source=python -m pytest
         coverage report
+        tach test
     - name: Check types with pyright
       run: |
         pyright --pythonversion ${{ matrix.python-version }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,15 +17,35 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
         check-latest: true
+
+    - name: Cache Python dependencies
+      uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+
+    - name: Cache Rust dependencies
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
         pip install -e ".[dev]"
+
     - name: Check ruff
       run: |
         ruff check
@@ -45,3 +65,4 @@ jobs:
       run: cargo fmt --all --check
     - name: Check Rust
       run: cargo check
+
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,11 +50,10 @@ jobs:
       run: |
         ruff check
         ruff format --check
-    - name: Test with tach/pytest and report coverage
+    - name: Test with pytest and report coverage
       run: |
         coverage run --branch --source=python -m pytest
         coverage report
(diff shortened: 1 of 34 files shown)

### CODE REVIEW

File: python/tach/cli.py [hunk 1]
```diff
@@ -532,7 +554,7 @@ def tach_sync(
             exclude_paths=exclude_paths,
         )
     except Exception as e:
-        print(str(e))
+        raise e
         sys.exit(1)
```
emdoyle, line 558: I think if we intend to simply re-raise `e` we should just eliminate the try/except entirely, and in my mental model the `sys.exit` won't ever be executed due to the re-raise anyway. But is there another motivation for this change?

python/tach/core/config.py
```diff
@@ -128,7 +137,7 @@ def add_dependency_to_module(self, module: str, dependency: str):
             self.modules.append(ModuleConfig(path=module, depends_on=[dependency]))
         else:
             # Config already exists, set the union of existing dependencies and new ones
-            new_dependencies = set(current_module_config.depends_on) | {dependency}
+            new_dependencies = set(current_module_config.depends_on) | {dependency}  # pyright: ignore[reportUnhashable]
```
emdoyle, line 140: Is Dependency actually hashable but pyright can't detect it?

python/tests/test_check.py
```diff
@@ -10,80 +9,12 @@
...
-                    is_end_of_path=True,
-                    full_path="domain_three",
-                    config=ModuleConfig(path="domain_three", strict=False),
-                    children={},
-                ),
-            },
-        )
-    )
+# from tests.conftest import *  # noqa
+# from tests.conftest import example_dir  # noqa
```
emdoyle, line 17: Are these intended to be commented out?

python/tach/sync.py
```diff
@@ -18,52 +19,50 @@ def sync_dependency_constraints(
...
+                    for module in project_config.modules
+                ]
+            }
+        )
+        # Update deprecations first
+        check_result = check(
+            project_root=project_root,
+            project_config=project_config,
+            exclude_paths=exclude_paths,
+        )
```
emdoyle, line 44: This is an expensive call, I think it'll noticeably slow down `sync` on larger codebases. If I understand correctly, the intention is to figure out which deprecated dependencies are actually causing warnings, and preserve those. Instead of running `check` here, could you instead record a mapping of the existing deprecated dependencies, and then add them back to the config if they appear as full-blown errors in the check below?

emdoyle (changes requested):
Looks good! Left some small comments and questions - only thing blocking is the second `check` call in `sync`. I could be missing something there though.

.github/workflows/ci.yml
```diff
@@ -17,15 +17,35 @@ jobs:
...
         python-version: ${{ matrix.python-version }}
         check-latest: true
+
+    - name: Cache Python dependencies
+      uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
```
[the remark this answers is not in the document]
caelean: Yes - It falls back on the most recent cache hit if there's not an exact hit. This is actually what we want, because we always run `pip install`, regardless of a cache hit or miss!

.github/workflows/ci.yml
```diff
@@ -45,3 +65,7 @@ jobs:
       run: cargo fmt --all --check
     - name: Check Rust
       run: cargo check
+    - name: Compile Maturin
+      run: maturin build
+    - name: Build Python
+      run: pip install .
```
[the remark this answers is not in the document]
caelean: whoops yeah leftover from testing

python/tach/core/config.py
caelean: Yep! Because of `model_config = {"frozen": True}`See https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.frozen

emdoyle (approved):
LGTM!

Issue #203, opened 2024-07-23
Support for PEP 420 Implicit Namespace Packages


malcolmgreaves:
Hello! 👋 Thank you for making and distributing `tach`! ❤️ 

Does `tach` support [namespace packages](https://peps.python.org/pep-0420/)? 🤔 I did not see explicit support in the docs. And I've been running into issues trying to adapt `tach` to my situation. I'd like to use `tach` in a project that makes use of multiple namespace packages. 

The individual packages are all PEP-420 compliant. There is a specific set of inter-dependencies that must be enforced between the namespaced packages. This DAG is currently encoded in each's own `install_requires`. However, I'd like to use `tach` to further enforce and check that these inter-namespace package dependencies are (a) all what we intend and (b) identify and flag any module boundary violations. It'd be great if `tach` would also inspect the `install_requires` and verify that the dependency graph is correct too. Here, "module" really means "distinct namespace package." Though I think the concepts are well aligned :)

Is this something that is in the near-future for `tach`?

malcolmgreaves:
Also, I think that this issue is quite related: https://github.com/gauge-sh/tach/issues/192 I did not want to distract from that individual's issue by bringing up explicit PEP-420 support. But perhaps a solution can work for both issues?

emdoyle:
This is now supported natively in `0.9.0`!
https://docs.gauge.sh/usage/configuration#source-roots

Issue #192, opened 2024-07-19
Support multiple source roots


vlaci:
In a monorepo, top-level packages are defined like this:

```
.
├── package_1_root
│   ├── package_1  # python source lives here
│   ... # possibly additional data directories
│   └── tests
├── package_2_root
│   ├── package_2 
│   └── tests
...
└── package_n_root
    ├── package_n 
    └── tests
```
The source is split into subdirectories, but the whole repository consists of a single python package. This layout can be achieved with e.g. `poetry`:

```toml
[tool.poetry]
name = "monorepo"
packages = [
    { include = "package_1", from = "package_1_root" },
    { include = "package_2", from = "package_2_root" },
#   ...
    { include = "package_n", from = "package_n_root" },
]
```

Then these packages can be used like this from python code:

```python
>>> import package_1.bar
>>> package_1.bar
<module 'package_1.bar' from '/.../package_1_root/package_1/bar/__init__.py'>
```

I need to set `source_root` to e.g. `package_1`, because otherwise `tach` thinks that the `package_<n>_root` is part of the module path.

I'd like to enforce dependencies in and between all of these `python_<n>` packages, so it would be great to be able to tell `tach` that individual packages can reside under different prefixes.

dhananjaypai08:
Wouldn't we much rather prefer package_1_root, package_2_root, .... package_n_root to be n different modules inside one single package folder that can act as a source root. Having multiple `<source_root>` will only become complex for the end developers using tach integrations as it would have n different tach.yml but on a broader scope this can include defining `<source_root>` in multiple packages but only `checking/syncing` those source roots that the user wants

emdoyle:
> In a monorepo, top-level packages are defined like this:
> 
> ```
> .
> ├── package_1_root
> │   ├── package_1  # python source lives here
> │   ... # possibly additional data directories
> │   └── tests
> ├── package_2_root
> │   ├── package_2 
> │   └── tests
> ...
> └── package_n_root
>     ├── package_n 
>     └── tests
> ```
> 
> The source is split into subdirectories, but the whole repository consists of a single python package. This layout can be achieved with e.g. `poetry`:
> 
> ```toml
> [tool.poetry]
> name = "monorepo"
> packages = [
>     { include = "package_1", from = "package_1_root" },
>     { include = "package_2", from = "package_2_root" },
> #   ...
>     { include = "package_n", from = "package_n_root" },
> ]
> ```
> 
> Then these packages can be used like this from python code:
> 
[...]

PR #208, opened 2024-07-29
Support multiple source roots and namespace packages

emdoyle:
Fixing #203 and #192 

NOTE: The base of this branch will be updated after #207 merges

This PR essentially changes `source_root` from a single path to a list of paths (`source_roots`).
This change propagates through most of the codebase, but the semantic change is straightforward.

Tach will now search for project imports in _all_ the configured source roots when translating module paths to filesystem paths (and vice versa).

This unlocks support for 'monorepo' use cases, since a user can now mark all applicable package source roots and validate the expected dependencies.

This PR also drops certain expectations around finding `__init__.py` in Python packages, which means [namespace packages](https://peps.python.org/pep-0420/) are also supported. This is particularly useful when building many subsidiary packages which are distributed under a common namespace.

Example:
image


README.md
```diff
@@ -147,6 +147,7 @@ python/tach/cache/setup.py[L7]: Import 'tach.filesystem.find_project_config_root
 Tach also supports:
 
 - [Manual file configuration](https://docs.gauge.sh/usage/configuration)
+- [Monorepos and namespace packages](https://docs.gauge.sh/usage/configuration#source-roots)
```
caelean, line 150: Small nit - thoughts on the ordering of the list here? Was trying to go from most important to least for the average user, almost put deprecations under inline exceptions. Imo this is prob under strict interfaces/maybe we want to move manual file configuration down too

docs/usage/configuration.mdx
```diff
@@ -57,48 +57,78 @@ Each module listed under the `modules` key above can accept the following attrib
...
-This means Tach will expect that your Python imports are resolved relative to the directory in which `tach.yml` exists (the 'project root').
+The `source_roots` key is required for Tach to understand the imports within your project.
+If it is not set explicitly, `source_roots` defaults to your project root path: `['.']`.
+This means Tach will expect that your Python imports are resolved relative to the directory in which `tach.yml` exists.
 
-Below is a typical case in which `source_root` is necessary.
+Below is a typical case in which modifying `source_roots` is necessary.
 
-### Example
+### Example: Monorepo
```
caelean, line 70: Should this replace the standard example, or should this be in addition to it? I could see monorepo/multiple source root configuration being it's own docs page as it's less likely to represent the path that most take

python/tach/core/config.py
```diff
@@ -83,15 +83,15 @@ class ProjectConfig(Config):
...
     exact: bool = False
     disable_logging: bool = False
     ignore_type_checking_imports: bool = True
     forbid_circular_dependencies: bool = False
 
-    @field_serializer("source_root")
-    def serialize_source_root(self, source_root: Path, _) -> str:
-        return str(source_root)
+    @field_serializer("source_roots")
+    def serialize_source_root(self, source_roots: list[Path], _) -> list[str]:
```
caelean:
```suggestion
    def serialize_source_roots(self, source_roots: list[Path], _) -> list[str]:
```

python/tach/mod.py
```diff
@@ -56,11 +60,14 @@ def validate_configuration(
...
-        if configuration.source_root not in module_path.parents:
+        if not any(
+            source_root in module_path.parents
+            for source_root in configuration.source_roots
+        ):
             # This module exists outside of the source root
             # This is not allowed and should be reported as a configuration error
             errors.append(
-                f"Module '{module_path}' is not contained within source root: '{configuration.source_root}'"
+                f"Module '{module_path}' is not contained within any source root: {configuration.source_roots}"
```
caelean, line 70:
```suggestion
                f"Module '{module_path}' is not contained within any source root: {[str(root) for root in configuration.source_roots]}"
```

caelean (approved):

caelean:
Looks great/working for me when I test! Some nice-to-haves would mainly be around the ergonomics of `mod` - unmarking modules that don't have a source root, not allowing nested source roots, not allowing modules with the same relative path under different source roots, etc. 

Definitely all nonblocking tho; stoked to see this get out!

README.md
emdoyle: Updated this :+1:

docs/usage/configuration.mdx
emdoyle: Good idea, restored the previous example

malcolmgreaves:
This is awesome! Just tried this today on my own project and it worked so well! Thank you so much :) ❤️

PR #220, opened 2024-07-30
Cycles detection is now binary

caelean:
Detecting and printing all cycles is an unsafely bound operation on large repositories. 

Because of this, we are reducing the scope of `forbid_circular_dependencies` to just detect if _any_ cycle exists for each module in your tach.yml.

In the future, we may introduce tooling around exposing the _first found_ cycle for a given module.

### DIFF
--- a/python/tach/parsing/modules.py
+++ b/python/tach/parsing/modules.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 import networkx as nx
+from networkx import NetworkXNoCycle
 
 from tach.core import ModuleTree
 from tach.errors import TachCircularDependencyError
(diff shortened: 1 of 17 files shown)


python/tach/parsing/modules.py
```diff
@@ -44,13 +45,21 @@ def find_cycles(
...
-
-    canonical_cycles = {tuple(canonical_form(cycle)) for cycle in all_cycles}
-
-    unique_cycles = [list(cycle) for cycle in canonical_cycles]
+    modules_with_cycles: list[str] = []
+    for module in modules:
+        module_path = module.path
+        try:
+            # Find *any* cycle, starting with module_path
+            cycle: list[tuple[str, str]] = nx.find_cycle(graph, source=module_path)  # type: ignore
```
caelean, line 53: `source` is confusing in `find_cycle` - it just starts with `source`, but moves on to other nodes if there's no cycle found with `source`. 

There's not a better method to call either unfortunately, so we need to check that the cycle found contains `source` after the fact. It does at least mean if we don't find any cycles on the first call we can just return.

emdoyle (approved):

PR #222, opened 2024-08-03
Add report-external command

emdoyle:
cc: @devstein

This PR introduces `report-external`, which detects usages of third-party packages in a given file path. It can also be used with `--raw` to generate machine-readable output suitable for piping into a `requirements.in` during a build process.

Example:
```
> tach report-external --raw python/tach
pyyaml
pydantic
stdlib-list
importlib-metadata
gitpython
prompt-toolkit
networkx
rich
```


devstein:
@emdoyle 🙇‍♂️

Issue #124, opened 2024-06-21
Support file globbing in exclude_paths


emdoyle:
As mentioned in #101, it's confusing that `exclude_paths` uses regex instead of file globbing.

Conceptually, this behavior should match that of other linters like ruff. Using that example, it makes sense to support `globset`:
https://docs.astral.sh/ruff/settings/#exclude
https://docs.rs/globset/latest/globset/#syntax

This will require updating the docs to explain the new behavior.

emdoyle:
> hey @emdoyle I would love to work on this issue. since, its good first issue 👍

Sounds great! Feel free to put up a PR any time, and ask me if anything is unclear.

emdoyle:
@shreyash2002 is this still something you'd like to work on? Otherwise I'll unassign so that someone else can pick it up.

legendkartik45:
Hi @emdoyle,

Is this issue still open? If so, could you please assign it to me? I would love to work on it over the weekend!

emdoyle:
@legendkartik45 That would be great, thanks! I might also recommend basing your branch on #208 to avoid significant merge conflicts when your changes are ready.

caelean:
Hey @legendkartik45! Are you still working on this issue?

caelean:
@legendkartik45 going to open this one back up for other folks

architgarg:
I tried fixing this issue here - https://github.com/gauge-sh/tach/pull/234

sparshg:
Resolved in #259

PR #234, opened 2024-08-06
Major Change: Support file globbing in exclude_paths

architgarg:
Fixes: https://github.com/gauge-sh/tach/issues/124

1. This PR updates the code to use glob paths instead of regex patterns for the exclude property.
2. Documentation has been updated accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -486,6 +486,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40ecd4077b5ae9fd2e9e169b102c6c330d0605168eb0e8bf79952b256dbefffd"
 
 [[package]]
+name = "glob"
+version = "0.3.1"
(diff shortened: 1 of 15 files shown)


architgarg:
@caelean I don't know Rust, but the code seemed simple, so I attempted to fix the issue without testing it locally. Could you test/review it?

emdoyle (changes requested):
Thanks for this! I pushed a handful of additional changes and it's now working well for me locally.

This will cause issues with existing config though, and I doubt we can auto-migrate regex to glob patterns since they are not as expressive. I think we should roll this out under a separate key (maybe `exclude_glob`) and deprecate the regex patterns while still accepting it for a period of time.

I should  be able to push those changes later, but feel free to take a stab at it if you get the time!

emdoyle:
Closing since the commits landed through #259 !

Issue #123, opened 2024-06-21
Project configuration should be parsed in Rust


emdoyle:
While profiling Tach during #82 , I noticed that `parse_project_config` was taking ~100ms, even though it should be parsing a relatively tiny file into a relatively simple data structure. Instead of trying to improve this in Python, it makes more sense to just extract this to the Rust extension and use something like `serde`.

Either in the same PR or a separate one, it would also make sense to move the `dump_project_config_to_yml` into the Rust extension as well.

Ideally all of this can be done in a way that is minimally coupled to YML, since we may want to migrate to TOML in the future.

ranjanmangla1:
@emdoyle can I take this one?

emdoyle:
@ranjanmangla1 This one is also open! Do you want me to assign this one as well as #145 ?

ranjanmangla1:
@emdoyle assign it, will love to take this one too

caelean:
@ranjanmangla1 curious as to if you're still working on this one?

caelean:
@ranjanmangla1 going to open this one back up for other folks

PR #235, opened 2024-08-07
Migrate from YAML to TOML, Python to Rust for configuration

emdoyle:
Fixes: #123 

This PR moves our configuration parsing to Rust, while maintaining the existing Python interface. This is a temporary situation while we work on porting the rest of the downstream code to Rust as well.

The primary user-facing change is that we are moving from YAML to TOML for configuration.

There is code in this PR that automatically migrates a YAML config file to TOML, in addition to the pre-existing auto-migrations for other field changes we've made. For almost all users the switchover should be automatic and transparent.


emdoyle:
Planning to update docs and README in a separate PR along with the version bump. It gives us some time after merging before we have to release.

PR #236, opened 2024-08-07
tach sync removes non-exixtent modules

### LINKED ISSUE #145: sync should delete modules which do not exist
To see this behavior, you can add:
```
- path: notarealmodule
  depends_on: []
...
```
into your `tach.yml`, and then run  `tach sync`.

Right now, the module `notarealmodule` will remain in the `tach.yml` file.
This PR should instead remove all modules listed which cannot be resolved from the configured source root.

tiluckdave:
Fixes #145


tiluckdave:
image

tiluckdave:
@emdoyle can you review

python/tach/sync.py
```diff
@@ -84,12 +84,26 @@ def sync_project(
...
+            module_path = fs.module_to_pyfile_or_dir_path(
+                tuple(project_config.source_roots), module.path
+            )
+            if module_path is not None:
+                existing_modules.append(module)
+            else:
+                print(f"Removing non-existent module: {module.path}")
+        project_config = ProjectConfig(
+            **project_config.model_dump(exclude={"modules"}), modules=existing_modules
+        )
```
emdoyle, line 99: I think it is better to combine this logic with the section on line 31 within sync_dependency_constraints. Using `model_copy` will be faster than dumping and revalidating, and it is easier to reason about the code when all of the prune-related config changes are next to each other.

emdoyle (changes requested):
A couple minor things, then this should be good to go. Thanks for doing this!

python/tach/sync.py
tiluckdave: @emdoyle I made the changes but that caused two `check` tests to fail as check compares the synced modules with the config and throws error for unused dependencies.

tiluckdave: Any thoughts. why that would happen?
I am not able to understand

tiluckdave: image

When I run `tach sync` and `tach check` inside valid example project. No erros are shown.

emdoyle: Thanks for the details, I pulled down the branch locally and was able to debug. Essentially all modules were being pruned out because `source_roots` is expected as absolute paths, meaning with relative paths from the test directory, none of the modules were being found. That also would explain why running it manually from the directory itself hides the issue :)

I made a small change so that the source roots are absolute and now tests are passing as expected.

emdoyle (approved):

python/tach/sync.py
tiluckdave: Got it.

PR #259, opened 2024-08-20
Continue #234

sparshg:
Rebased to main and created an abstraction for pattern matching.

### DIFF
--- a/python/tach/constants/__init__.py
+++ b/python/tach/constants/__init__.py
@@ -7,13 +7,7 @@ PACKAGE_FILE_NAME = "package"
 ROOT_MODULE_SENTINEL_TAG = "<root>"
 TACH_YML_SCHEMA_URL = "https://raw.githubusercontent.com/gauge-sh/tach/v0.10.0/public/tach-yml-schema.json"
 
-DEFAULT_EXCLUDE_PATHS = [
-    "**/tests/**",
-    "**/docs/**",
-    "**/venv/**",
-    "**/__pycache__/**",
-    "**/*egg-info/**",
-]
+DEFAULT_EXCLUDE_PATHS = ["tests", "docs", "venv", ".*__pycache__", ".*egg-info"]
 
 __all__ = [
     "PACKAGE_NAME",
--- a/python/tach/constants/__init__.py
+++ b/python/tach/constants/__init__.py
@@ -7,7 +7,13 @@ PACKAGE_FILE_NAME = "package"
 ROOT_MODULE_SENTINEL_TAG = "<root>"
 TACH_YML_SCHEMA_URL = "https://raw.githubusercontent.com/gauge-sh/tach/v0.10.0/public/tach-yml-schema.json"
 
-DEFAULT_EXCLUDE_PATHS = ["tests", "docs", "venv", ".*__pycache__", ".*egg-info"]
+DEFAULT_EXCLUDE_PATHS = [
+    "**/tests/**",
+    "**/docs/**",
+    "**/venv/**",
+    "**/__pycache__/**",
+    "**/*egg-info/**",
+]
 
 __all__ = [
     "PACKAGE_NAME",
--- a/python/tach/interactive/modules.py
+++ b/python/tach/interactive/modules.py
@@ -160,7 +160,7 @@ class FileTree:
                     # Exclude patterns are relative to project root, and may include a trailing slash
                     if exclude_paths is not None and use_regex_matching is None:
                         raise errors.TachError(
-                            f"Must specify whether to use regex matching when providing exclude paths."
+                            "Must specify whether to use regex matching when providing exclude paths."
                         )
 
                     entry_path_for_glob = str(entry.relative_to(self.root.full_path))
--- a/python/tach/interactive/modules.py
+++ b/python/tach/interactive/modules.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import re
+import fnmatch
 from collections import deque
 from dataclasses import dataclass, field
 from enum import Enum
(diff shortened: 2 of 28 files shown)


sparshg:
imo we can have `pattern_matching = "globs"` by default added in config and if someone wants to use regex, they can change it to `pattern_matching = "regex"`

caelean:
@sparshg  if there's only going to be two types of pattern_matching, I think a `bool` is sufficient.

I'd also prefer `glob` to be the default, but we don't want to break any existing installations. I think if we're writing a brand new `tach` config it's fine to set it to `glob`, but if we detect or are working with an existing config, we should not break the existing behavior.

Because of this, it might be easier to just default to `regex` for now, and create a better migration mechanism in the future. What do you think?

sparshg:
Ready to merge, can deprecate if `use_regex_matching` is detected in config, in the next major release.

python/tach/interactive/modules.py
```diff
@@ -119,14 +120,18 @@ def build_from_path(
         path: Path,
         depth: int | None = 1,
         exclude_paths: list[str] | None = None,
+        use_regex_matching: bool | None = None,
```
caelean, line 123:
```suggestion
        use_regex_matching: bool | None = False,
```
Let's explicitly default to False for now

python/tach/interactive/modules.py
```diff
@@ -152,11 +158,19 @@ def _build_subtree(
                         continue
 
                     # Exclude patterns are relative to project root, and may include a trailing slash
-                    entry_path_for_matching = (
-                        f"{entry.relative_to(self.root.full_path)}/"
-                    )
+                    if exclude_paths is not None and use_regex_matching is None:
```
caelean, line 161: We shouldn't break existing configs.

Ideally this pr simply provides to option for users to add the flag and turn it on if they want to use globs, but simply upgrading the package should not result in an error if you already have excludes.

python/tests/test_sync.py
```diff
@@ -1,15 +1,24 @@
...
 def test_valid_example_dir(example_dir, capfd):
     project_root = example_dir / "valid"
-    with pytest.raises(SystemExit) as exc_info:
-        # TODO not a great test, can change the example fs
-        tach_sync(project_root=project_root, add=True)
-    assert exc_info.value.code == 0
-    captured = capfd.readouterr()
-    assert "✅" in captured.out  # success state
+
+    with tempfile.TemporaryDirectory() as temp_dir:
```
caelean, line 15: Thank you 🙏

python/tach/constants/__init__.py
```diff
@@ -7,7 +7,13 @@
 ROOT_MODULE_SENTINEL_TAG = "<root>"
 TACH_YML_SCHEMA_URL = "https://raw.githubusercontent.com/gauge-sh/tach/v0.10.0/public/tach-yml-schema.json"
 
-DEFAULT_EXCLUDE_PATHS = ["tests", "docs", "venv", ".*__pycache__", ".*egg-info"]
+DEFAULT_EXCLUDE_PATHS = [
```
caelean, line 10: The default exclude paths should match the default of `use_regex_matching` in the project config. Given that, I don't think we should update this yet.

python/tach/interactive/modules.py
```diff
@@ -135,6 +140,7 @@ def _build_subtree(
         root: FileNode,
         depth: int = 1,
         exclude_paths: list[str] | None = None,
+        use_regex_matching: bool | None = None,
```
caelean, line 143:
```suggestion
        use_regex_matching: bool | None = True,
```
Let's default to true for now

python/tach/interactive/modules.py
caelean: One thing we could do is validate that if you have `is_regex_matching` as `True`, is that all of the `exclude_paths` are valid regex. Similarly for if it's `False` and `globs`!

caelean (changes requested):
Looking great, just a few requests on behavior for existing users!

Could we add error handling and a  test for globbing vs. regex checking as well?

python/tach/constants/__init__.py
sparshg: True, missed this from older commits

python/tach/interactive/modules.py
sparshg: The default value of True if being passed into this function by Rust (where all other defaults are also being set), I don't think we should do it specifically here.

python/tach/interactive/modules.py
sparshg: It won't break the existing config, since the config is loaded with `ProjectConfig.with_derived_unset_fields`, and as the old configs won't have `use_regex_matching` field, it will default it automatically to True.
In short, this `if` condition is never meant to run for the end user. If it does, its our mistake somewhere

sparshg: Pattern validation is a nice idea, will add soon.

sparshg:
I have made some changes, `--exclude` parameter now works. `exclude_paths` parameter gets extended with `project_config.exclude` only inside `cli.py` and also gets validated there.

caelean:
@sparshg thank you! Let me know when you're ready for another review and I'll do some testing and we can get this deployed 🔥

sparshg:
Ready to review, 

in 2e17a87 I had tried to fix `--exclude` by adding cli excludes to project config. 

62cf8a5 reverts this decision, because excludes from project config should be imo immutable. Also, did a fixup to revert some other bad changes in 760f507

lmk what you think about current approach.

sparshg:
The pattern validation is not done in `tach_check_external()`. It does not take `--exclude` from cli, is this meant to be there? I will handle validation accordingly then.

caelean (approved):

PR #278, opened 2024-08-28
Fix parsing for pypoetry deps. 

### LINKED ISSUE #273: tach check external does not seem to parse dependencies with extras correctly
In my repo, in my `pyproject.toml` I have a dependency on pandera with extras:

`"pandera[strategies]>=0.18.3",`

Running `tach check-external` reports "Undeclared dependencies" for the places where we use `pandera`, presumably because the extras are not properly parsed?

sparshg:
Fixes #273


PoulavBhowmick03:
Thanks man! Appreciate it

PR #280, opened 2024-08-28
sarif support

### LINKED ISSUE #247: Support code-version-system reporting through SARIF and/or rdjson output formats
In order to take advantage of a CVS's ability to highlight issues in their UI (i.e Github review/annotation), producing output in a format that standard CVS reporters (i.e. Reviewdog) can ingest will be very useful.

There are two output formats that are generic for static analysis output which are SARIF and rdjson. SARIF is a good choice for as wide a range of CVS reporters as possible and is not tied to any particular CVS reporter. However, rdjson (specific to Reviewdog) is also very good additional format to support as Reviewdog can produce very detailed reviews/annotations for the main CVS-es from such.

caelean:
Adds a `--sarif` flag to `tach check`which additionally outputs the error results to a minimal `tach-check-results.sarif` file.

Fixes #247.


malcolmrebughini:
@caelean was there a particular reason why this PR never made it? I'm looking into taking the tach output to show in github PRs and into SonarCloud.

caelean:
Hey @malcolmrebughini - this ended up getting deprioritized for us. If it's particularly important to you, I can look at getting it back on the roadmap!

PR #284, opened 2024-08-31
Porting core to Rust

sparshg:
WIP

This PR aims to slowly port the tach's python core to rust.

- [x] Move AST handling completely to Rust
- [x] Ported some `ModuleTree` and `ModuleNode` code
- [x] Port `check_import` and its dependent functions
- [x] Port `check`
- [x] Port `sync`
- [x] Port configs
- [x] Fix toml string formatting (For now the returned toml string from rust, is read again with `tomli` in python before writing to file)
- [x] Port `test`
- [x] Delete `core/modules.py` and `parsing/modules.py`
- [x] Port tests (Partially done)
- [x] Delete all ported python counterparts (Probably more to delete)


sparshg:
There are probably a lot of bugs, since this is the first attempt to replace the `check` command with the ported version. I have ported other things required by this in top down manner.

sparshg:
I can't believe the port just ran as expected after adding over 800 lines, with just one logical mistake, rust is great, if it compiles it runs!
ofc still some bugs exist, will fix them soon.

sparshg:
@emdoyle Apologies for such a big PR, I have ported many tests to Rust, and all others are passing except `test_addected_modules.py` since it needs to be ported too. (`tach test` works as before).
Current version is stable and can be reviewed, although I need to do some more work, probably port `test.py` to make all tests pass.

Oh and I compared, `time tach check` on main branch takes around `0.34 ms`, but on this branch takes `0.25 ms`. (tested on the same codebase ofcourse). So about 36% faster!

src/core/module.rs
```diff
@@ -0,0 +1,291 @@
...
+        &mut self,
+        config: ModuleConfig,
+        path: String,
+        interface_members: Vec<String>,
+    ) -> Result<(), ModuleTreeError> {
+        if path.is_empty() {
+            return Err(ModuleTreeError::InsertNodeError);
+        }
+
+        let mut node = Rc::get_mut(&mut self.root).unwrap();
```
emdoyle, line 121: If it's easy, I'm just now realizing we'll probably want to use `Arc` instead. Once this PR lands and we're stable with essentially the whole `tach check/sync` in Rust we should make the import checking concurrent (each file can be processed completely independently).

But no pressure to make changes here, just making a note that this will likely change in the future.

src/parsing/module.rs
```diff
@@ -0,0 +1,112 @@
...
+    // Find strongly connected components (SCCs)
+    let sccs = kosaraju_scc(&graph);
+
+    // Filter SCCs to find cycles
+    let mut modules_with_cycles = Vec::new();
+    for scc in sccs {
+        if scc.len() > 1 {
+            modules_with_cycles.extend(scc);
+        }
+    }
```
emdoyle, line 51: Very nice - do you have a feel for the runtime cost of this?

emdoyle (commented):
Absolutely awesome.

@sparshg Thanks so much for these contributions! I left some comments, but mainly just making sure I understand the details in some areas.

I am also happy to jump into the remaining tasks you mention in the PR description - I'm going to start developing from this branch as a base anyway since it doesn't make sense to write more in Python-land when this is on the horizon.

src/parsing/module.rs
sparshg: kosaraju runs in `O(V+E)` (#Vertices + #Edges). The python version finds cycle using DFS starting on every node, and checks the cycle if the node exists in cycle which should be `O(VE(V+E))` loosely. Yet to delete the python implementation.

src/check_int.rs
```diff
@@ -0,0 +1,405 @@
...
+    Filesystem(#[from] fs::FileSystemError),
+    #[error("Module tree error: {0}")]
+    ModuleTree(#[from] parsing::error::ModuleTreeError),
+    #[error("Exclusion error: {0}")]
+    Exclusion(#[from] exclusion::PathExclusionError),
+}
+
+#[derive(Debug, Clone)]
+#[pyclass(get_all, module = "tach.extension")]
+
```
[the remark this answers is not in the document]
sparshg: Yep it does

src/core/module.rs
sparshg: Its very easy, we can just find/replace `Rc` with `Arc` (I had it in Arc a few commits ago), so I guess its better to do in future when concurrent checking is being implemented.

python/tests/test_report_external.py
```diff
@@ -4,24 +4,24 @@
...
-            "src/pack-b/src",
-            "src/pack-c/src",
-            "src/pack-d/src",
-            "src/pack-e/src",
-            "src/pack-f/src",
-            "src/pack-g/src",
-        ],
-        ignore_type_checking_imports=True,
-    )
+    p = ProjectConfig()
```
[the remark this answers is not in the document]
sparshg: There is no constructor for `ProjectConfig` that accepts those arguments in Rust, I would have to explicitly make one for this, so I decided to set fields manually here for now.

sparshg:
@emdoyle This can be merged, more tests can be ported, and there are some `TODO`s in the code, but maybe for another PR, this is already long enough.

src/core/module.rs
sparshg: Changed to `Arc`

emdoyle (approved):
Working smoothly for me! Makes sense to merge this and clean up incrementally so that we can review smaller PRs :sweat_smile: 

Will hold off on shipping a new release from main for a little while so that this can stabilize, and we'll bump a minor version (0.11.0). Also going to hard-deprecate the regex exclude matching.

Thanks again!

PR #285, opened 2024-08-31
Better error handling

sparshg:
This PR is to support and prepare #284 for the restructure.
Uses `thiserror` module to restructure and make error handling more consistent.

Also handles a critical bug of infinite recursion and crash due to `err.into()` in Line 50
https://github.com/gauge-sh/tach/blob/b53cb3e45ffe1e428300d45832c4b0311c7c423e/src/lib.rs#L46-L50


emdoyle (approved):
Awesome, thanks for this!

PR #286, opened 2024-09-01
Able to generate mermaid locally save in a '.md' and image in '.png' file. Open to change if required.

### LINKED ISSUE #276: `tag show` export to mermaid locally
It would be practical if we can export locally to Mermaid in addition to graphviz `.dot`. I see the web version (`tach show --web`) already renders to Mermaid. This would make it possible to auto-update diagrams in a README.md for example.

There also exist tools to convert dot to mermaid: https://github.com/monitorian/dot2mermaid

tosoham:
Fixes #276

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,11 @@ dependencies = [
     "stdlib-list>=0.10.0; python_version < '3.10'",
     "eval-type-backport>=0.2.0; python_version < '3.10' and python_version > '3.7'",
     "importlib_metadata>=6.5; python_version == '3.7'",
+    "code2flow>=2.5.1",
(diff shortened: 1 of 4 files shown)


emdoyle:
Hi @tosoham , thanks for taking a stab at this issue. Unfortunately I'm closing this PR since it's not really what the issue is looking for. I can give a few pieces of feedback on it so that you are aware:
- the dependencies added are very heavy and would require many end users to separately install GraphViz to install Tach
- even though this is meant to render locally, this sends data to `mermaid.ink` which would violate the user's expectation
- this appears not to use the Tach module graph at all, instead pulling in `code2flow` which generates callgraphs

PR #309, opened 2024-09-10
Restore `TachCircularDependecyError` pretty print

sparshg:
Regression introduced during porting to Rust

Before

```
Module tree error: Circular dependency detected: ["domain_one", "domain_two", "domain_three"]
```

After

```
❌ Circular dependency detected for module 'domain_one'
❌ Circular dependency detected for module 'domain_two'
❌ Circular dependency detected for module 'domain_three'

Resolve circular dependencies.
Remove or unset 'forbid_circular_dependencies' from 'tach.toml' to allow circular dependencies.
```


emdoyle (approved):
Nice catch

PR #323, opened 2024-09-20
Update cli.py

pillont-dh:
Hi ! 
I use your lib with github runner, but I have a problem during the display of the result in this environment : 

ERROR : UnicodeEncodeError: 'charmap' codec can't encode character '\u2705' in position 0: character maps to <undefined>

So I propose this update to work also in github display ! \o/

Thanks for your work, this tools is verry usefull ! 💪

### DIFF
--- a/python/tach/cli.py
+++ b/python/tach/cli.py
@@ -58,7 +58,7 @@ def build_error_message(error: BoundaryError, source_roots: list[Path]) -> str:
         )
 
     error_template = (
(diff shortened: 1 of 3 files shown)


emdoyle:
Thanks for alerting us to this issue! I think these changes make sense, but I would prefer them being optional, or even auto-detecting whether the display might support the extended unicode. I'll look into this and see if I can make some changes to this branch.

emdoyle (approved):

PR #474, opened 2024-12-10
Move to abi3 wheels

emdoyle:
[Docs](https://pyo3.rs/v0.23.3/building-and-distribution.html#building-python-extension-modules)

This should dramatically reduce our footprint on PyPI by building only one wheel for each OS/Arch combo instead of building one for every version of Python.

This also stops building wheels for PyPy for now, since opting into abi3 automatically stops building interpreter-specific wheels, but this can be restored if requested. For now since we don't know if these are being used, we should err on the side of not building wheels since we recently hit the [PyPI size limit](https://github.com/pypi/support/issues/5184).


.github/workflows/publish.yml
```diff
@@ -34,19 +29,12 @@ jobs:
...
-            3.11
-            3.12
-            3.13
+          python-version: 3.13
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
           target: ${{ matrix.platform.target }}
-          args: --release --out dist -i python3.7 python3.8 python3.9 python3.10 python3.11 python3.12 python3.13 pypy3.7 pypy3.8 pypy3.9 pypy3.10
+          args: --release --out dist --find-interpreter
```
emdoyle, line 37: Most likely, find-interpreter now does nothing, but leaving this in to see if it will pick up PyPy automatically.

Issue #579, opened 2025-01-27
VSCode extension - tach.toml not found


eugenenelou:
Hi, 
I'm trying to setup the VSCode extension and I get this error `tach.toml not found`.

The problem may be that my python project is not at the root of my repository. but I only found options to change where `tach` is imported from, but not about `tach.toml`.

emdoyle:
Thanks for raising this - we are working on fixing this now! https://github.com/gauge-sh/tach-vscode/pull/19 & #581

emdoyle:
@eugenenelou This is now possible with Tach >= `0.25.2` and the VSCode extension == `0.25.3`! There is a new setting available (see screenshot below). The path should include the `tach.toml` stem itself, and actually allows any valid TOML file (even with another name).

Please let us know if this works for you!  Changes are thanks to @max-muoto

![Image](https://github.com/user-attachments/assets/71243269-4803-4ddc-ae40-b4bac89120ef)

eugenenelou:
It works perfectly, thanks a lot !

PR #581, opened 2025-01-29
feat: Support custom `toml` path

max-muoto:
This PR adds support for custom `tach.toml` paths for all `tach server` for usage in the VSCode plugin.


python/tach/cli.py
```diff
@@ -21,17 +20,7 @@
     TachSetupError,
     TachVisibilityError,
 )
-from tach.extension import (
```
max-muoto, line 24: Avoiding direct imports for functions so we can directly mock without  including the `cli` module in the path for the mocked function. When you import directly like this instead of mocking `tach.extension.my_func` you mock the `cli` module which can be a little bit awkward. The Google Python style guide additionally recommends this:

> Use import statements for packages and modules only, not for individual types, classes, or functions.

Let me know if we're fine with this!

python/tach/cli.py
```diff
@@ -1003,18 +1008,28 @@ def current_version_is_behind(latest_version: str) -> bool:
...
     except Exception as e:
         print(f"Failed to parse project config: {e}")
         sys.exit(1)
 
 
-def main() -> None:
-    args, parser = parse_arguments(sys.argv[1:])
+def main(argv: list[str] = sys.argv[1:]) -> None:
+    args, parser = parse_arguments(argv)
     project_root = fs.find_project_config_root() or Path.cwd()
```
emdoyle, line 1025: Given the use case in #579 , we also need to adjust the `project_root` which is passed to `tach_server` based on the location of the actual config file. This is needed  for Tach to correctly interpret `source_roots` and differentiate between first-party and third-party imports.

I believe the config parsing code in Rust tracks the location of the config file, but it would be better for now to just adjust the usage of `find_project_config_root` in `main` to handle this specific LSP case.
and `tach server`.

python/tach/cli.py
```diff
@@ -1003,18 +1008,28 @@ def current_version_is_behind(latest_version: str) -> bool:
...
-        return parse_project_config(project_root)
+        return parse_project_config(project_root, file_name=file_name)
     except Exception as e:
         print(f"Failed to parse project config: {e}")
         sys.exit(1)
 
 
-def main() -> None:
-    args, parser = parse_arguments(sys.argv[1:])
+def main(argv: list[str] = sys.argv[1:]) -> None:
```
emdoyle, line 1023: Not a huge deal because direct CLI usage is unlikely, but in this function there is a call to `print_no_config_found` which should now take the custom config argument so that the error message can be more accurate. Right now it will always say `tach.toml` is not found, even when specifying another name.

max-muoto: Fixed: https://github.com/gauge-sh/tach/pull/581/commits/2b9ccf4d1444618116e937b1de213465038a2ded

python/tach/cli.py
max-muoto: Should have caught this, fixed: https://github.com/gauge-sh/tach/pull/581/commits/2b9ccf4d1444618116e937b1de213465038a2ded

Issue #585, opened 2025-02-01
make `tach check` print more information about layer violations


crockeo:
`CodeDiagnostic` only includes information about the `tach` modules where an issue occurs, not the Python module where they occur. This means `tach check` can't include all of the information which would allow a developer to identify a problem import without also opening the file in which it occurs.

As an anonymized example for a `LayerCheckResult`:

```
❌ path/to/python/file.py[L3]: Cannot import 'path.to.module'. Layer 'infrastructure' ('path.to.python') is lower than layer 'product' ('path.to.module').
```

I would then have to open `path/to/python/file.py` and see that the import in question is maybe something like:

```python
from path.to.module.enums import SomeEnum
```

I think it would be useful for `tach check` to include information about both:

- The Python module name of the file which contains the bad import (although this is easily inferred if you assume `PYTHONPATH=.`).
- The Python module name of the imported module.

I'm happy to take a whack at it if y'all are ok with including this behavior.

emdoyle:
I agree, the diagnostic should include the actual import module path! The intended behavior IIRC is for that first path (`path.to.module` in your example error) to be the full import path (e.g. `path.to.module.enums.SomeEnum`).

The Python module name of the file is less of an obvious need to me, since in general Tach's diagnostics try to reference module paths which appear in the config, and the filepath is clickable to get directly to the referenced line.

We'd definitely appreciate a PR here! I'm in the middle of a refactor (#586) but I should be able to easily transplant the fix -- might be another day or two before this merges.

crockeo:
Fixed in #587. Thanks for merging it!

PR #587, opened 2025-02-02
use correct `import_mod_path` for `check_layers`

crockeo:
Addresses https://github.com/gauge-sh/tach/issues/585.

It looks like this function was using `target_module_config.path` instead of `import_mod_path` to populate `import_mod_path` for the violations. I think this is the right solution, because from testing locally `import_mod_path` seems to be the full module + attribute name of the thing being imported.

Let me know if y'all are ok to merge this, or if you'd prefer to wait until [the refactor](https://github.com/gauge-sh/tach/pull/586) is done.

Also let me know if there's anything you'd like me to change. I didn't get around to setting up tests in local dev (complaining about missing pyo3 symbols), so it may need some work 😅

### DIFF
--- a/src/commands/check/checks.rs
+++ b/src/commands/check/checks.rs
@@ -24,7 +24,12 @@ fn check_dependencies(
     layers: &[String],
 ) -> Result<(), Diagnostic> {
     // Layer check should take precedence over other dependency checks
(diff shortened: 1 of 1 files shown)


emdoyle (approved):
Appreciate the fix!

PR #616, opened 2025-02-12
Migrate check external interface fully to Rust

max-muoto:
This PR migrates `check_external` fully to Rust, in particular the pulling in of relevant metadata so that we can easily fix https://github.com/gauge-sh/tach-vscode/issues/17.


emdoyle (approved):
Looks great!

PR #618, opened 2025-02-13
Add check external to LSP server

max-muoto:
This PR adds `tach-check-external` to the LSP server. Additionally it adds the `auto-initialize` feature so that a Python interpreter is automatically initialized when we use `with_gil`.

Fixes: https://github.com/gauge-sh/tach-vscode/issues/17


src/commands/check/check_external.rs
```diff
@@ -100,7 +102,63 @@ impl<'a> FileChecker<'a> for CheckExternalPipeline<'a> {
...
 }
 
-pub fn check(
+struct CheckExternalMetadata {
+    module_mappings: HashMap<String, Vec<String>>,
+    stdlib_modules: Vec<String>,
+}
+
+/// Get metadata for checking external dependencies.
+fn get_check_external_metadata(
```
max-muoto, line 111: Ended up pushing this down one layer deeper.

src/commands/check/check_external.rs
```diff
@@ -100,7 +102,63 @@ impl<'a> FileChecker<'a> for CheckExternalPipeline<'a> {
...
+        Ok(CheckExternalMetadata {
+            module_mappings,
+            stdlib_modules,
+        })
+    })
+}
+
+pub fn check(project_root: &Path, project_config: &ProjectConfig) -> Result<Vec<Diagnostic>> {
+    let metadata = get_check_external_metadata(project_config)?;
+    check_with_modules(
```
max-muoto, line 153: From my perspective having a non-public interface like this for testing is a simpler strategy than mocking, but let me know what you think.

emdoyle: Definitely agree, particularly in Rust. The boilerplate of stubbing out mock implementations alone is annoying, and testing the 'real thing' gives more confidence. In Python I still sometimes use mocks since monkey patching is so easy.

src/lsp/server.rs
```diff
@@ -165,17 +180,16 @@ impl LSPServer {
...
-                        return e.into();
-                    }
-                }
-                None
-            })
+        let check_external_result = check_external(&self.project_root, &self.project_config)?;
+
+        let check_diagnostics = self.filter_diagnostics_results(check_result, &uri_pathbuf);
+        let check_external_diagnostics =
+            self.filter_diagnostics_results(check_external_result, &uri_pathbuf);
```
emdoyle, line 187: Just noticed, we should be able to make the filter function generic with a trait bound for `IntoIterator<Item = Diagnostic>`, then apply it after chaining the results together as owning iterators. But definitely not a blocker for this PR!

emdoyle (approved):
Looks great! Glad to get this shipped.

PR #620, opened 2025-02-16
Support ignore patterns from `.gitignore`

max-muoto:


dav3yblaz3 (approved):
ok for test thanks team

emdoyle (approved):
Appreciate the initial implementation on this! I ended up going with a slightly different approach - replacing our `walkdir` usage with the bundled `ignore` file walking APIs. Also saw an opportunity to centralize this into a `filesystem` struct which handles exclude paths too. After this lands we should fully remove `use_regex_matching` and delegate exclude path globs to `ignore` as well.

max-muoto:
> Appreciate the initial implementation on this! I ended up going with a slightly different approach - replacing our `walkdir` usage with the bundled `ignore` file walking APIs. Also saw an opportunity to centralize this into a `filesystem` struct which handles exclude paths too. After this lands we should fully remove `use_regex_matching` and delegate exclude path globs to `ignore` as well.

Awesome, thanks Evan!

PR #634, opened 2025-02-19
contributing docs troubleshooting

### LINKED ISSUE #621: Tach dev installation does not support python3 compatible command
hi folks, 
 I was trying to setup tach using the contributing docs. I am using a `mac` based machine and my machine only supports `python3` commands  and not `python` commands. 
I tried aliasing python3 to python and then perform `make deps` but it does not find python as a valid command(Alias doesn't work in non interactive environment i guess)

my fix : 
- creating a permanent symlink via this command : `sudo ln -s $(which python3) /usr/local/bin/python`

Could this also help other developers setting up tach?(I could help in contributing to the troubleshooting doc)

dhananjaypai08:
This PR fixes #621 

Additionally:
- The already existing contributing docs needed some markdown fixing
- Troubleshooting for machine based setups
- Most of the changes for step2 to build tach


docs/contributing/setting-up-project.mdx
```diff
@@ -2,36 +2,102 @@
...
+```powershell
+python -m venv venv
+venv\Scripts\Activate
+```
+
+### Step 2: Install Dependencies
+
+```bash
+make install
+```
```
emdoyle, line 39: I think this section should be removed, since `make deps` is already suggested in the preceding section. Does `make deps` work on your system? Under the hood it installs and uses `uv` to create and manage a virtual environment in `.venv`, so the developer should not need to manage this manually.

docs/contributing/setting-up-project.mdx
```diff
@@ -2,36 +2,102 @@
...
  make deps
 ```
 
-#### 2. Build
-To build and rebuild after changes to Rust files
+## 2. Build
+
+To build and rebuild after changes to Rust files.
+
+> **Note:** Make sure you have the Rust compiler installed. This package requires Rust and Cargo to compile extensions.
```
emdoyle, line 15: Mintlify allows tags like `<Note>` which should look nicer, e.g.

```suggestion
<Note>
Make sure you have the Rust compiler installed. This package requires Rust and Cargo to compile extensions.
</Note>
```

docs/contributing/setting-up-project.mdx
```diff
@@ -2,36 +2,102 @@
...
+sudo ln -s $(which python3) /usr/local/bin/python
 ```
 
-#### 3. Test
+#### Linux Fix:
+On Linux, you may need to install the `python-is-python3` package or create a symlink manually:
+
+```bash
+sudo apt install python-is-python3  # For Debian/Ubuntu
+```
```
emdoyle, line 59: I am suspicious of installing a package for this, and would prefer not to recommend it to downstream users.

emdoyle (commented):
Thanks for contributing! I added some suggested changes I think we should make before merging.

docs/contributing/setting-up-project.mdx
dhananjaypai08: I used `make deps` while setting up and got few errors : 1. maturin not found 2. `uv` didn't create a venv for `tach` to work

docs/contributing/setting-up-project.mdx
dhananjaypai08: yes agreed

docs/contributing/setting-up-project.mdx
```diff
@@ -2,36 +2,80 @@
...
+
+```powershell
+Set-Alias -Name python -Value python3
+```
+
+After applying the appropriate fix, retry:
+
+```bash
+make install
+```
```
emdoyle, line 61: This section should actually be next to `make deps`, and should refer to `make deps` instead of `make install`. From looking at the Makefile, that command is the one that actually invokes `python` directly.

docs/contributing/setting-up-project.mdx
emdoyle: I think those errors must have come from `make install` rather than `make deps`. I left a separate comment but I think this `python` issue is likely related to that command instead.

emdoyle (approved):
Thanks for the changes!

PR #648, opened 2025-02-26
Add layers to report upload / show

caelean:


src/modularity/diagnostics.rs
```diff
@@ -40,6 +42,8 @@ impl TryFrom<Diagnostic> for UsageError {
             Some(member),
             Some(usage_module),
             Some(definition_module),
+            usage_layer,
+            definition_layer,
```
emdoyle, line 46: Pushed a small change to omit the `Some` in this match pattern - otherwise any Diagnostic without layers would fail to match and be discarded. One thing to note is that this means `null` is actually going to be passed in `usage_layer` and `definition_layer`, in case that has implications for the platform downstream.

emdoyle (approved):
LGTM!

Issue #614, opened 2025-02-12
tach show flag to hide deprecated connections


skloibhofer:
I am at the moment refactoring a project, and during this process I marked quite some dependencies as deprecated.
To show what the future dependency graph should look like, I would like to have a flag to hide (or mark) the deprecated connections.

I know I could make a new tach.toml with something like `grep -v deprecated tach.toml`, but I think this is a common use case, that would warrant a new flag. Something like 
 
```
  --hide-deprecated     Hide deprecated connections from the dependency graph.
```

If you agree, and the flag can be implemented in the python part of the project, I could try to submit a PR for this.

emdoyle:
Hi @skloibhofer , this should be possible to implement in Python at the moment, since `show` still operates primarily in Python (although this may change soon).

I would be happy to review a PR for this feature, let me know if you are still open to this!

skloibhofer:
Nice, I try to submit a PR next week!

emdoyle:
@skloibhofer That makes sense. I am realizing now that the logic for `show --web` has diverged pretty significantly from the local `show`. For this feature, it is probably best to implement it for local mermaid and graphviz, but ignore it and print a warning if using `--web`, since that responsibility is held by the remote UI.

skloibhofer:
Took me longer than expected, but I added a first minimal PR now. At the moment this only changes the edge style to dashed (or dotted for mermaid, as I found no dashed style), to resemble the web view. 

Possible further enhancements / changes:
* Use `build_modules` instead of `project_config.filtered_modules` also for dot_file and mermaid functions. This would standardize the used data with web, and should have the same functionality
* Add a `--hide-deprecated` flag as initially suggested. Marking the deprecated dependencies is enough for me atm, but this could still be an enhancement
* Should I add a minimal test for this feature?

emdoyle:
Going to close this issue since #700 is now released! To your questions:

- yes we should consolidate the logic, but want to take the opportunity to move the functions to Rust as well, although not a high priority at the moment
- if this behavior works for you, let's hold off on this, or can track it in another issue
- already merged after manual testing, seemed quite low risk and straightforward!

PR #700, opened 2025-03-21
Mark deprecated dependencies in git show (offline)

skloibhofer:
As discussed in #614 
First commits only changed edge type to dashed / dotted for deprecated dependencies.


emdoyle (approved):
LGTM! This works when I test the resulting DOT/Mermaid files in viewers.

Issue #697, opened 2025-03-19
Incorrect version number


bevanjkay:
Hi, Homebrew maintainer here 👋,

While regression testing the latest version, it looks like the version number was not correctly incremented to `0.28.2`.

```
    Minitest::Assertion: Expected /0\.28\.2/ to match "tach 0.28.1\n".
```

Homebrew ref: https://github.com/Homebrew/homebrew-core/pull/211660

emdoyle:
This will be resolved in `0.28.3` which will be released shortly!

bollwyvl:
Appears to have occurred again on `0.28.5`.

PR #701, opened 2025-03-21
fix: patch version to match with the release

chenrui333:
fixes https://github.com/gauge-sh/tach/issues/697
relates to https://github.com/Homebrew/homebrew-core/pull/211660

cc @emdoyle


python/tach/__init__.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import annotations
 
-__version__: str = "0.28.1"
```
emdoyle, line 3: This is the one that backs `tach --version`

emdoyle (approved):

python/tach/__init__.py
chenrui333: I see, thanks! :)

Issue #696, opened 2025-03-19
tach falsely identifies external dependencies as undeclared when compiling dependencies from `pyproject.toml` -> `requirements.txt`


cameronbrill:
In this case, `requirements.txt` contains dependencies of the dependencies in `pyproject.toml`. It's the dependency diff (i.e. those that are in `requirements.txt` but not in `pyproject.toml`) that is being flagged by `tach check-external`. I imagine this would likely be an issue for projects that use [uv](https://docs.astral.sh/uv/)'s `uv.lock` file or [poetry's](https://python-poetry.org/) `poetry.lock` as well (but that's not what my team is blocked by).

emdoyle:
Appreciate the patience on this issue! If I understand correctly, you are saying that `tach check-external` emits errors when you use packages which are only declared in your compiled `requirements.txt`, but not in your `pyproject.toml`.

If you are distributing this package to other users or orgs with the `pyproject.toml` file, then I would actually recommend declaring all dependencies in `pyproject.toml` if they are ever imported in your code, rather than relying on the transitive dependencies to be stable. This is the use case that Tach assumes.

But if you are just building/deploying this package internally and are using `requirements.txt` as the source of truth for that purpose, I can see why this should take precedence over `pyproject.toml`. This might generally the case when both are present, so I'm okay with making `requirements.txt` always take precedence over `pyproject.toml` for dependency declarations. Will put this into the next release.

DetachHead:
closing this as per the comments above

PR #720, opened 2025-03-31
Make `requirements.txt` take precedence over `pyproject.toml` for external dep checks

emdoyle:
Fixes: #696


emdoyle:
Closing for now, this doesn't seem like the right solution

PR #723, opened 2025-04-01
fix: bump version to match with 0.28.5 release

chenrui333:
👋 patch the version info again :)

https://github.com/Homebrew/homebrew-core/pull/217509


emdoyle:
:man_facepalming: my bad this happened again!

IIUC this change will make it to homebrew due to the patch pointing to this commit?

emdoyle (approved):

PR #799, opened 2025-06-17
require repo for git ignores

altendky:
I have a repo rooted at my home directory.  I have a `~/.gitignore` with `**/**` in it.  I have my git clones mostly located under `~/repos/tach` and similar.  tach was looking beyond the git clone repos and finding the `~/.gitignore` and ignoring all files instead of analyzing them.  This seems to still respect the 'local' `.gitignore` files while not processing the `~/.gitignore` that is beyond the root of the project work tree.


altendky:
and yes, i did see the note about being unmaintained.  just figured i'd put this on the record regardless.

Issue #854, opened 2025-11-19
document that tach can be configured in `pyproject.toml`


DetachHead:
Transferred from: https://github.com/DetachHead/dtach/issues/16

as far as i can tell the docs only mention `tach.toml`, but it can be configured using `pyproject.toml` as well

PR #883, opened 2025-12-08
Add support for dependency groups

mluogh-xdof:
References https://github.com/tach-org/tach/issues/565

Add basic support for including dependency groups when using `check-external`. When the bool is set to true, the set of declared dependencies includes everything in all the dependency groups.

This should be a no-op by default as it is `default` (false). 

If desired, later on we can do something more advanced like what the FR requests, but this would be enough for my use right now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "tach"
-version = "0.32.2"
+version = "0.33.0"
 authors = [
     { name = "Caelean Barnes", email = "caeleanb@gmail.com" },
     { name = "Evan Doyle", email = "evanmdoyle@gmail.com" },


src/external/parsing.rs
```diff
@@ -95,6 +108,59 @@ fn extract_deps_from_value(dependencies: &mut HashSet<String>, deps: &Value) {
...
+
+    let Some(group_value) = groups.get(group_name) else {
+        return;
+    };
+
+    let Some(group_array) = group_value.as_array() else {
+        return;
+    };
+
+    const EXCLUDED_DEPS: [&str; 3] = ["python", "poetry", "poetry-core"];
```
DetachHead: idk about hardcoding specific dependencies like this
- would `python` itself ever even be declared in a dependency group? in my projects (that use uv) the python version is specified using `requires-python`
- poetry (and other similar tools such as uv) shouldn't be installed in dependency groups because they are managed by the tool itself right? or does poetry do it differently? it's been years since i've used poetry so i might be misremembering how it works

src/external/parsing.rs
```diff
@@ -195,3 +261,147 @@ pub fn parse_requirements_txt(requirements_path: &Path) -> Result<HashSet<String
...
+        let content = r#"
+[project]
+name = "test"
+
+[dependency-groups]
+test = ["pytest", {include-group = "nonexistent"}]
+"#;
+        let file = create_temp_pyproject(content);
+        // Should not fail, just skip the missing group
+        let result = parse_pyproject_toml(file.path(), true).unwrap();
```
DetachHead: it should fail imo

src/external/parsing.rs
mluogh-xdof: I am a UV user myself, so also not familiar with poetry. 

My understanding from reading similar code in this file is that poetry will declare these as dependencies under 
[tool.poetry.dependencies] causing failures if you run it with `--exact` since those dependencies are not used in code.

However I guess it wouldn't exist in the [dependency-groups]

public/tach-toml-schema.json
```diff
@@ -249,6 +249,18 @@
...
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of module:name pairs to rename imports (e.g. 'PIL:pillow')"
+        },
+        "include_dependency_groups": {
+          "type": "boolean",
+          "default": false,
+          "description": "Include dependencies from PEP 735 [dependency-groups] in pyproject.toml"
```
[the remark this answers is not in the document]
mluogh-xdof: That should be under external (starts at L243)

public/tach-toml-schema.json
```diff
@@ -249,6 +249,18 @@
...
+        },
+        "rename": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of module:name pairs to rename imports (e.g. 'PIL:pillow')"
+        },
+        "include_dependency_groups": {
+          "type": "boolean",
```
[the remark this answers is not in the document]
mluogh-xdof: To support this, I have changed it to read from pyproject.toml instead of a single global file, to better support the uv workspace monorepo pattern.

mluogh-xdof:
I believe many of these issues should be fixed now! PTAL

docs/usage/configuration.md
```diff
@@ -586,6 +586,30 @@ In most cases you should not need to specify `rename` manually (see the Note bel
...
+# In pyproject.toml
+[dependency-groups]
+dev = ["ruff", "mypy"]
+test = ["pytest", "coverage"]
+
+[tool.tach.external]
+include_dependency_groups = ["dev", "test"]
+```
+
+This configuration is read from each package's `pyproject.toml`, making it suitable for monorepos where different packages may have different dependency groups.
```
DetachHead: this kind of implies that this is the only part of tach configurable in `pyproject.toml`. i think to be consistent with the rest of the docs it should mention `tach.toml`. there is a separate issue to document that tach can be configured via `pyproject.toml` though (#854)

src/external/parsing.rs
```diff
@@ -110,14 +127,25 @@ fn extract_deps_from_value(dependencies: &mut HashSet<String>, deps: &Value) {
 
 /// Extracts dependencies from PEP 735 [dependency-groups] table.
 /// Handles both direct package names and {include-group = "..."} references.
+/// If `include_groups` contains "all", all groups are processed.
+/// Otherwise, only the specified groups are processed.
```
DetachHead: i think to avoid ambiguity with a dependency group called `"all"`, we should only allow `"all"` as a string instead of an item in the array. ie. the python type would be `list[str] | Literal["all"]`.

this will also prevent users from specifying `"all"` with other dependency groups, which would be redundant

PR #885, opened 2025-12-12
fix: pytest plugin now skips all tests in unaffected files

andre-vauvelle:
## Summary

Fixes a bug where only the first test from each unaffected file was skipped, instead of all tests in that file.

## The Bug

When running `tach test --base HEAD` (no changes), only 1 test per file was being skipped:

```
collected 8 items
[Tach] Skipped 1 test file (1 tests) since they were unaffected by current changes.
============================== 7 passed ===============================
```

## Root Cause

In `pytest_collection_modifyitems`, after processing the first test from a file:
1. The path was added to `seen`
2. Subsequent tests from the same file hit `if item.path in seen: continue`
3. This skipped the removal logic, leaving those tests in the collection

## The Fix

Use a two-pass approach:
1. **First pass**: Identify all paths that should be removed (without modifying the list)
2. **Second pass**: Filter out all items from those paths at once using list comprehension

## After Fix

```
collected 8 items
[Tach] Skipped 1 test file (8 tests) since they were unaffected by current changes.
============================ no tests ran =============================
```

## Testing

Tested with a minimal example project:
- No changes: All 8 tests correctly skipped ✓
- With changes to module: All 8 tests correctly run ✓

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -1543,6 +1543,14 @@
                 "code": "reportUnusedParameter",
                 "range": {
                     "startColumn": 4,
+                    "endColumn": 11,
+                    "lineCount": 1
+                }
+            },
+            {
+                "code": "reportUnusedParameter",
+                "range": {
+                    "startColumn": 4,
                     "endColumn": 14,
                     "lineCount": 1
                 }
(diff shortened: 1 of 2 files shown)


python/tach/pytest_plugin.py
```diff
@@ -66,29 +65,30 @@ def pytest_collection_modifyitems(
...
+                handler.remove_test_path(item.path)
+
+    # Second pass: remove all items from paths marked for removal
+    original_count = len(items)
+    items[:] = [
+        item for item in items
+        if not item.path or item.path not in paths_to_remove
+    ]
+    handler.num_removed_items = original_count - len(items)
+
```
DetachHead: i think we can simplify this by using a `pytest_collect_file` hook instead:

```py
class HasTachConfig(Protocol):
    config: TachConfig

@pytest.hookimpl(wrapper=True)
def pytest_collect_file(
    file_path: Path, parent: HasTachConfig
) -> Generator[None, list[pytest.Collector], list[pytest.Collector]]:
    handler = parent.config.tach_handler
    # skip any paths that already get filtered out by other hook impls
    result = yield
    if result and handler.should_remove_items(file_path=file_path.resolve()):
        handler.num_removed_items += 1
        handler.remove_test_path(file_path)
        return []
    return result
```

DetachHead (commented):
thanks for the contribution!

andre-vauvelle:
Thanks for maintaining this project and for the great review suggestions! I've implemented the `pytest_collect_file` hook approach as you recommended - it's much cleaner than the two-pass `pytest_collection_modifyitems` approach.

Also added a `--tach-validate` mode and some documentation for standalone plugin usage in a separate commit.

DetachHead (approved):
looks good! will merge when the CI passes 🚀

DetachHead:
fyi just pushed a small change becasue i want to avoid baselining new errors ([more info here](https://docs.basedpyright.com/latest/benefits-over-pyright/baseline/#how-often-do-i-need-to-update-the-baseline-file))

PR #893, opened 2025-12-29
feat: improve pytest plugin UX with auto-loading and smart defaults

andre-vauvelle:
## Summary

Improves the pytest plugin UX based on feedback from #885:

- **Auto-loads via pytest11 entrypoint** - no need for `-p tach.pytest_plugin`
- **Default: reports only, doesn't skip** - shows what could be skipped without changing test outcomes
- **Only skips when `--tach-base` is provided** - explicit opt-in to skipping
- **Auto-detects default branch** (main/master)
- **Duration caching** - estimates time saved/could be saved
- **Validation is implicit** - warns if would-be-skipped tests fail

## Usage

```bash
# Default: runs all tests, shows what could be skipped
pytest
# Output: [Tach] 42 tests in 8 files unaffected by changes (~15.3s could be saved). Skip with: pytest --tach-base main

# Skip unaffected tests
pytest --tach-base main
# Output: [Tach] Skipped 8 test files (42 tests) (~15.3s saved) - unaffected by current changes.

# Detailed output
pytest --tach-verbose

# Disable plugin
pytest --no-tach
```

## New Options

- `--tach-base`: Base commit to compare against. When provided, unaffected tests are skipped.
- `--tach-head`: Head commit to compare against (default: current filesystem)
- `--tach-verbose`: Show detailed output including changed files and skipped test paths
- `--no-tach`: Disable the tach pytest plugin entirely

## Depends on

#885

### DIFF
--- a/docs/usage/commands.md
+++ b/docs/usage/commands.md
@@ -356,7 +356,7 @@ The plugin auto-detects whether your default branch is `main` or `master`.
 
 - `--tach`: Enable test skipping using the auto-detected base branch
 - `--tach-base <commit>`: Set the base commit explicitly (also enables skipping)
-- `--tach-head <commit>`: Head commit to compare against (default: current filesystem)
+- `--tach-head <commit>`: Head commit to compare against (also enables skipping. default: current filesystem)
 - `--tach-verbose`: Show detailed output including changed files and skipped/would-skip test paths
 
 To disable the plugin entirely, use pytest's built-in plugin disabling:
--- a/python/tach/pytest_plugin.py
+++ b/python/tach/pytest_plugin.py
@@ -110,7 +110,7 @@ class TachPluginState:
     """True if --tach or --tach-base was explicitly provided to enable skipping."""
     verbose: bool
     base: str
-    head: str
+    head: str | None
     would_skip_paths: set[Path]
 
 
@@ -217,7 +217,7 @@ def pytest_configure(config: Config):
 
     tach_flag = cast("bool", config.getoption("--tach"))
     tach_base_option = cast("str | None", config.getoption("--tach-base"))
-    head = cast("str", config.getoption("--tach-head")) or ""
+    head = cast("str | None", config.getoption("--tach-head")) or None
     verbose = cast("bool", config.getoption("--tach-verbose"))
 
     # Skipping is enabled if --tach, --tach-base, or --tach-head is provided
@@ -434,10 +434,8 @@ def pytest_report_collectionfinish(
 
 def pytest_terminal_summary(
     terminalreporter: TerminalReporter,
-    exitstatus: int,
     config: Config,
 ):
-    _ = exitstatus  # unused, required by pytest hook signature
     # Check if plugin is active
     if tach_state_key not in config.stash:
         return
--- a/python/tach/filesystem/git_ops.py
+++ b/python/tach/filesystem/git_ops.py
@@ -143,7 +143,7 @@ def get_current_branch_info(
 
 
 def get_changed_files(
-    project_root: Path, head: str = "", base: str = "main"
+    project_root: Path, head: str | None = None, base: str = "main"
 ) -> list[Path]:
     # Local import because git-python takes ~80ms to load
     from git import GitCommandError, InvalidGitRepositoryError, NoSuchPathError, Repo


andre-vauvelle:
The difference is you get it in one run with actionable output. Running pytest then tach test separately means you have to repeat tests and manually cross-reference which failures were in skipped files - validate just tells you directly `test_foo::test_bar failed and would have been skipped`. Also useful as a scheduled CI job to monitor impact analysis accuracy without someone having to manually compare two runs each time, doing this manually would not scale either.

DetachHead:
thinking about this some more, i think it might make more sense to always output this validation info when running pytest normally, otherwise users are unlikely to discover this feature and remember to use it when running pytest locally. perhaps we could refactor the plugin to work like so:

- the plugin is active by default when tach is installed (using the [`pytest11` entrypoint](https://docs.pytest.org/en/stable/how-to/writing_plugins.html#making-your-plugin-installable-by-others)). this will make it more consistent with how other pytest plugins work
- by default, the plugin does not skip any tests. it just runs them normally and also outputs these validation results
- tests are only skipped when specifying `--tach-base` or `--tach-head` arguments
- since this could be considered a breaking change, `tach test` (which is the publicly documented api) should probably behave the same as it does now. we can then consider deprecating it in the future in favor of the pytest plugin (this isn't the only reason i want to deprecate `tach test`, see #894)

andre-vauvelle:
Thanks for the feedback! I've refactored based on your suggestions:

**Changes:**

1. **Plugin auto-loads via pytest11 entrypoint** - no need for `-p tach.pytest_plugin`
2. **Default behavior: runs all tests, reports what could be skipped**
   ```
   [Tach] 42 tests in 8 files unaffected by changes (~15.3s could be saved). Skip with: pytest --tach-base main
   ```
3. **Only skips tests when `--tach-base` is explicitly provided**
4. **Auto-detects default branch** (main/master)
5. **Added `--tach-verbose`** for detailed output
6. **Added `--no-tach`** to disable the plugin entirely
7. **Duration caching** - shows estimated time saved/could be saved
8. **Validation is now implicit** - if a would-be-skipped test fails in default mode, shows a warning

This means users get impact analysis info for free just by having tach installed, and can opt-in to skipping tests with `--tach-base`.

docs/usage/commands.md
```diff
@@ -330,6 +330,55 @@ Using `pytest`, running `tach test` will perform [impact analysis](https://marti
...
+```
+
+The plugin auto-detects whether your default branch is `main` or `master`.
+
+**Options:**
+
+- `--tach-base`: Base commit to compare against. When provided, unaffected tests are skipped. (default: auto-detects main/master)
+- `--tach-head`: Head commit to compare against (default: current filesystem)
+- `--tach-verbose`: Show detailed output including changed files and skipped/would-skip test paths
+- `--no-tach`: Disable the tach pytest plugin entirely
```
DetachHead: i don't think we need a `--no-tach` argument, since pytest plugins can be disabled using the `-p` flag:

```
pytest -p no:tach
```

we should document this though, and also document that you can also disable the plugin in `pyproject.toml` (see https://docs.pytest.org/en/stable/how-to/plugins.html#deactivating-unregistering-a-plugin-by-name)

python/tach/pytest_plugin.py
```diff
@@ -1,48 +1,187 @@
...
+        try:
+            result = subprocess.run(
+                ["git", "rev-parse", "--verify", branch],
+                cwd=project_root,
+                capture_output=True,
+            )
+            if result.returncode == 0:
+                return branch
+        except Exception:
+            pass
```
DetachHead: this could probably also use a check to ensure that both `main` and `master` don't exist... which i've seen happen oin some repos before

python/tach/pytest_plugin.py
```diff
@@ -1,48 +1,187 @@
...
-        raise TachSetupError("In Tach pytest plugin: No project config found")
+        # No tach config found, silently disable
+        return
 
-    base = config.getoption("--tach-base")
+    tach_base_option = config.getoption("--tach-base")
     head = config.getoption("--tach-head")
 
+    # Track if skipping is enabled (--tach-base explicitly provided)
+    config.tach_skip_enabled = tach_base_option is not None
```
DetachHead: thinking about this some more, if the skipping functionality is enabled using the `--tach-base` argument, that means the automatic detection of the main branch can never be used right? because the user would always have to explicitly specify a branch name.

maybe it makes sense to create a separate argument to enable the skipping functionality, like just `--tach`. but that argument is only needed if `--tach-base` or `--tach-head` aren't explicitly specified.

for example:

- `pytest --tach` - determines the default branch automatically and enables skipping functionality
- `pytest --tach-base master` - sets the default branch name explicitly and enables skipping functionality

docs/usage/commands.md
```diff
@@ -330,6 +330,55 @@ Using `pytest`, running `tach test` will perform [impact analysis](https://marti
...
+
+```bash
+pytest
+```
+
+By default, the plugin runs all tests but reports how many could be skipped based on impact analysis. This allows you to see the potential benefit without committing to skipping tests:
+
+```
+[Tach] 42 tests in 8 files unaffected by changes (~15.3s could be saved). Skip with: pytest --tach-base main
+```
```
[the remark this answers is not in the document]
andre-vauvelle: Agreed!

docs/usage/commands.md
andre-vauvelle: Good idea, I removed the flag, updated the test and added it to the docs.

docs/usage/commands.md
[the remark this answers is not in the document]
andre-vauvelle: added `To disable this message: pytest -p no:tach`

docs/usage/commands.md
```diff
@@ -330,6 +330,72 @@ Using `pytest`, running `tach test` will perform [impact analysis](https://marti
...
+
+```bash
+pytest -p no:tach
+```
+
+You can also disable it in `pyproject.toml`:
+
+```toml
+[tool.pytest.ini_options]
+addopts = "-p no:tach"
```
DetachHead: nit: this is the old config format. as of pytest 9 toml is properly supported so it can be configured like this:
```suggestion
[tool.pytest]
addopts = ["-p",  "no:tach"]
```

python/tach/pytest_plugin.py
```diff
@@ -16,46 +16,47 @@
 
 TACH_DURATIONS_CACHE_KEY = "tach/durations"
 
-# ANSI color codes
-_COLORS_ENABLED: bool | None = None
+# Rich console for colored output (uses same settings as tach.console)
+_console = Console(highlight=False, force_terminal=True)
```
DetachHead: this isn't exactly the same as the one in `tach.console` since we are specifying `force_terminal=True`. could maybe use a comment explaining why that's needed

python/tach/pytest_plugin.py
```diff
@@ -1,67 +1,260 @@
...
+    skip_enabled: bool
+    """True if --tach or --tach-base was explicitly provided to enable skipping."""
+    verbose: bool
+    base: str
+    head: str
+    would_skip_paths: set[Path]
+
+
+# StashKey for storing plugin state on pytest Config
+tach_state_key: StashKey[TachPluginState] = StashKey()
```
DetachHead:
```suggestion
tach_state_key: StashKey[TachPluginState] = StashKey()
"""`StashKey` for storing plugin state on pytest `Config`"""
```

python/tach/pytest_plugin.py
```diff
@@ -1,67 +1,260 @@
...
+
+
+def _save_durations(config: Config, durations: dict[str, float]) -> None:
+    """Save test durations to pytest cache."""
+    try:
+        cache = config.cache
+        if cache is not None:
+            cache.set(TACH_DURATIONS_CACHE_KEY, durations)
+    except Exception:
+        pass
```
DetachHead: what possible exceptions are we catching here? these functions don't look like they can fail because it safely handles `None` values

andre-vauvelle:
Addressed all review comments + CI passing now

python/tach/pytest_plugin.py
```diff
@@ -1,67 +1,270 @@
...
     project_root = fs.find_project_config_root() or Path.cwd()
     project_config = parse_project_config(root=project_root)
     if project_config is None:
-        raise TachSetupError("In Tach pytest plugin: No project config found")
+        # No tach config found, silently disable
+        return
+
+    tach_flag = cast("bool", config.getoption("--tach"))
+    tach_base_option = cast("str | None", config.getoption("--tach-base"))
+    head = cast("str", config.getoption("--tach-head")) or ""
```
DetachHead: nit: we should probably handle `None` consistently here. we should also change the default value for `head` in `get_changed_files` to `None` instead of an empty string
```suggestion
    tach_base_option = cast(str | None, config.getoption("--tach-base"))
    head = cast(str | None, config.getoption("--tach-head"))
```

python/tach/pytest_plugin.py
```diff
@@ -433,8 +433,11 @@ def _format_changed() -> str:
 
 
 def pytest_terminal_summary(
-    terminalreporter: TerminalReporter, _exitstatus: int, config: Config
+    terminalreporter: TerminalReporter,
+    exitstatus: int,
+    config: Config,
 ):
+    _ = exitstatus  # unused, required by pytest hook signature
```
DetachHead: pytest hook functions don't actually require all the fixture arguments. you can just remove the `exitstatus` argument completely and it'll still work

PR #898, opened 2026-01-01
add layer explict depends_on flag

### LINKED ISSUE #897: Option to explicitly enforce depends_on between layers
From the docs: 

> may freely depend on modules in lower layers, without declaring these dependencies

We'd like to always define `depends_on` even across layers, without it things like `tach show` don't show a true representation of the dependency graph. 

WDYT about adding a `tach.toml` flag for enforcing `depends_on` is always set? eg `layers_explicit_depends_on`?

craigmulligan:
Fixes #897

### DIFF
--- a/src/checks/internal_dependency.rs
+++ b/src/checks/internal_dependency.rs
@@ -121,7 +121,9 @@ impl<'a> InternalDependencyChecker<'a> {
             LayerCheckResult::Ok => {
                 // If layers_explicit_depends_on is enabled and target is not a utility,
                 // don't return early - fall through to check explicit depends_on declaration
-                if !self.project_config.layers_explicit_depends_on || dependency_module_config.utility {
+                if !self.project_config.layers_explicit_depends_on
+                    || dependency_module_config.utility
+                {
                     return Ok(vec![]);
                 }
                 // Fall through to check forbidden deps and depends_on


chadrik:
Hi, I'm interested in a feature like this, but I have a few questions:

- with `layers_explicit_depends_on=True` is `depends_on` intended to point at other modules or at other layers?  In the example the module name and the layer name are the same so it's hard to tell
- with `layers_explicit_depends_on=True` will `tach sync` populate  `depends_on`  with depended-on modules as it does when layers are not used?

I've just started testing out `tach` and I like the layer enforcement, but I don't like that when I use layers I lose the ability to visualize all of the dependencies between my modules with `tach show`.   IIUC, that's the goal of this PR, is that correct?

In an ideal world, the results of `tach show` would color-code module nodes based on layer and/or place them into [clusters](https://graphviz.org/Gallery/directed/cluster.html).  If this PR solves the core issue of recording all deps between modules whilst using layers, I might look into creating a PR for adding layer visualization.

DetachHead:
> - with `layers_explicit_depends_on=True` is `depends_on` intended to point at other modules or at other layers? In the example the module name and the layer name are the same so it's hard to tell

other modules. it should function the same as usual, but with the layers being enforced in addition. this feature is discussed in more detail at https://github.com/tach-org/tach/issues/897#issuecomment-3709753149

> - with `layers_explicit_depends_on=True` will `tach sync` populate `depends_on` with depended-on modules as it does when layers are not used?

it should do, since this PR doesn't touch any `tach sync` code. there should probably be a test that ensures it doesn't delete layer config though

craigmulligan:
Hey @DetachHead, could you take another look when you have a moment? 🙏

docs/usage/configuration.md
```diff
@@ -23,6 +23,8 @@ This is the project-level configuration file which should be in the root of your
 
 `forbid_circular_dependencies` (default: **false**) is a flag which causes `tach check` to fail if any circular dependencies are detected.
 
+`layers_explicit_depends_on` (default: **false**) is a flag which requires modules to explicitly declare dependencies on other layers using `depends_on`, even when those dependencies would be allowed by the layer hierarchy. By default, modules in higher layers can freely depend on modules in lower layers without declaring them. When this flag is enabled, all cross-layer dependencies must be explicit. [Utility modules](configuration.md#modules) remain accessible without explicit declaration.
```
DetachHead: this could probably use a link to the layers documentation

docs/usage/layers.md
```diff
@@ -22,6 +22,44 @@ When a module is assigned to a layer, this module:
...
+path = "user_repo"
+layer = "data"
+depends_on = []
+```
+
+With this flag enabled:
+
+- Modules must explicitly declare **all** cross-layer dependencies in `depends_on`
+- The layer hierarchy still prevents lower layers from depending on higher layers
+- [Utility modules](configuration.md#modules) remain accessible without explicit declaration
```
DetachHead: could use a note here that `depends_on` still needs to be explicitly specified for dependencies in the same layer (https://docs.gauge.sh/usage/layers/#how-does-it-work:~:text=must%20explicitly%20declare%20dependencies%20in%20its%20own%20layer)
```suggestion
- Modules must still explicitly declare dependencies in the same layer
- [Utility modules](configuration.md#modules) remain accessible without explicit declaration
```

DetachHead (commented):
thanks for your patience. just a few more nitpicky suggestions

.basedpyright/baseline.json
```diff
@@ -1,6 +1,34 @@
...
+        "./python/tach/cache/__init__.py": [
+            {
+                "code": "reportMissingTypeStubs",
+                "range": {
+                    "startColumn": 5,
+                    "endColumn": 22,
+                    "lineCount": 1
+                }
+            }
+        ],
```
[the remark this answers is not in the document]
craigmulligan: Makes sense, I've removed the baseline updates and left it for you to tweak 👍

DetachHead:
done

PR #908, opened 2026-01-27
fix crash when pytest `cacheprovider` plugin is disabled

tnewman-candid:
when pytest is invoked with `-p no:cacheprovider`, the cache attribute will not exist on the config object. This results in an attribute error. 

With this change, cache will be `None` in this case, allowing the rest of the duration functionality to continue under the case it was expected to.

see: https://github.com/tach-org/tach/issues/907


DetachHead:
looks good, do you mind resolving the conflicts so i can merge it

DetachHead:
was just a conflicting import. all good i just resolved it myself

PR #925, opened 2026-04-10
fix: resolve item.path to absolute in pytest_collection_modifyitems

andre-vauvelle:
## Summary

`pytest_collection_modifyitems` passes `item.path` directly to the `should_remove_items`, but `item.path` can be relative (e.g. when tests live in subdirectories like `tests/`). The Rust side only matches against absolute paths, so changed test files get incorrectly deselected — tach thinks they're unaffected.

**Fix:** call `item.path.resolve()` before passing to `should_remove_items`.

### Why existing tests didn't catch this

The existing tests use `pytester` which creates all test files at the project root. When files are at the root, relative and absolute paths are equivalent, so `item.path` already matches. The bug only manifests when tests are in subdirectories (the common real-world layout).

A `pytester` subprocess also always provides absolute `item.path` values, so even adding subdirectory integration tests doesn't reproduce the issue. The added unit test uses mocks to verify `resolve()` is called, which does fail without the fix.

## Test plan

- [x] Existing tests pass (`uv run pytest python/tests/test_pytest_plugin.py`)
- [x] New unit test fails without fix, passes with fix

### DIFF
--- a/python/tach/pytest_plugin.py
+++ b/python/tach/pytest_plugin.py
@@ -325,7 +325,9 @@ def pytest_collection_modifyitems(config: Config, items: list[Item]):
     if state is None:
         return
     items_to_remove = [
-        item for item in items if state.handler.should_remove_items(file_path=item.path)
+        item
+        for item in items
+        if state.handler.should_remove_items(file_path=item.path.resolve())
     ]
     state.handler.num_removed_items = len(items_to_remove)
 
--- a/python/tests/test_pytest_plugin.py
+++ b/python/tests/test_pytest_plugin.py
@@ -1,6 +1,11 @@
 from __future__ import annotations
 
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
 import pytest
+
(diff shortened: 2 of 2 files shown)


python/tach/pytest_plugin.py
```diff
@@ -325,7 +325,9 @@ def pytest_collection_modifyitems(config: Config, items: list[Item]):
     if state is None:
         return
     items_to_remove = [
-        item for item in items if state.handler.should_remove_items(file_path=item.path)
+        item
+        for item in items
+        if state.handler.should_remove_items(file_path=item.path.resolve())
```
DetachHead, line 330: > `pytest_collection_modifyitems` passes `item.path` directly to the `should_remove_items`, but `item.path` can be relative (e.g. when tests live in subdirectories like `tests/`).

would you mind providing a minimal repro of this issue? i can't seem to reproduce it myself. the tach repo itself has its tests located in a subdirectory (`python/tests`) and `item.path` seems to always be absolute no matter how i run it.

PR #930, opened 2026-05-08
fix: Support PEP 695 type aliases in interfaces

### LINKED ISSUE #929: Support PEP 695 type alias
After applying Ruff's UP040 rule (which upgrades code to the new Python 3.12 [PEP 695](https://peps.python.org/pep-0695/) type parameter syntax), `tach lint` now fails with something like:

`<filename>: TACH003: Interface '<app.path.Interface>' is unsupported type: TypeAlias`

It looks like Tach doesn't yet recognize the new type statement / generic type alias syntax introduced. I'm happy to contribute a fix for this — I already have a working patch for the AST ready and will open a PR shortly. Just wanted to file the issue first so it's tracked.

Steps to reproduce:
* Use a `type` alias with type parameters (PEP 695 syntax).
* Run `tach lint`.

chetmancini:
Fix for issue #929 

Recognize [PEP 695](https://peps.python.org/pep-0695/) `type Alias = ...` statements as exported module interface members. Reuse the existing annotation extraction path so aliases, annotated assignments, and function annotations keep the same primitive-type behavior. Add a regression test for a primitive PEP 695 alias.

First contribution -- let me know if I should add anything more testing or hygiene wise!

### DIFF
--- a/src/interfaces/data_types.rs
+++ b/src/interfaces/data_types.rs
@@ -78,6 +78,14 @@ pub struct InterfaceMember {
     node: InterfaceMemberNode,
 }
 
+fn annotation_name(annotation: &Expr) -> Option<String> {
+    match annotation {
+        Expr::Name(name) => Some(name.id.to_string()),
+        Expr::StringLiteral(s) => Some(s.value.to_string()),
+        _ => None,
+    }
+}
+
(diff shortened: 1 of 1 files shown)


DetachHead:
closing this, see https://github.com/tach-org/tach/issues/929#issuecomment-4412187021

Issue #853, opened 2025-11-19
duplicated code in `tach check` and `tach check-external`


DetachHead:
Transferred from: https://github.com/DetachHead/dtach/issues/29

https://github.com/DetachHead/dtach/blob/63c4a1d4217d4960bdcf726c7363282f1ffdbf71/src/commands/check/check_internal.rs#L189-L244

https://github.com/DetachHead/dtach/blob/63c4a1d4217d4960bdcf726c7363282f1ffdbf71/src/commands/check/check_external.rs#L231-L286

this should be extracted into a function

PR #931, opened 2026-05-09
fix: report syntax errors as errors, not warnings

### LINKED ISSUE #845: syntax errors are reported as a warning and do not effect the exit code
Transferred from: https://github.com/DetachHead/dtach/issues/27

```
me@pc:~/projects/test$ tach check
Configuration
⚠️  Warning: Skipped 'foo/baz.py' due to an unknown error.
⚠️  Warning: No first-party imports were found. You may need to use 'tach mod' to update your Python source roots. Docs: https://detachhead.github.io/dtach/usage/configuration#source-roots
✅ All modules validated!
me@pc:~/projects/test$ echo $?
0
```

### LINKED ISSUE #846: syntax errors are reported as an "unknown error" instead of a syntax error in `tach check`
Transferred from: https://github.com/DetachHead/dtach/issues/26

```
me@pc:~/projects/test$ tach check
Configuration
⚠️  Warning: Skipped 'foo/baz.py' due to an unknown error.
⚠️  Warning: No first-party imports were found. You may need to use 'tach mod' to update your Python source roots. Docs: https://detachhead.github.io/dtach/usage/configuration#source-roots
✅ All modules validated!
```

the correct error message is reported in `tach check-external`:
```
me@pc:~/projects/test$ tach check-external
Configuration
⚠️  Warning: Skipped 'foo/baz.py' due to a syntax error.
✅ All external dependencies validated!
```

kimjune01:
Fixes #845, fixes #846.

Syntax errors during `tach check` were reported as warnings and didn't affect the exit code. The `DiagnosticError::PythonParse` variant was missing from the error match arm, and `ImportParse` errors were routed to `new_global_warning` instead of `new_global_error`.

Both variants now produce `new_global_error`, so syntax errors surface as errors and fail the check.

### DIFF
--- a/src/commands/check/check_internal.rs
+++ b/src/commands/check/check_internal.rs
@@ -255,3 +255,66 @@ pub fn check(
 
     Ok(diagnostics)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::config::ProjectConfig;
+    use crate::diagnostics::Severity;
+    use crate::tests::fixtures::example_dir;
+    use rstest::*;
+
+    #[fixture]
+    fn multi_package_config() -> ProjectConfig {
+        use crate::config::ModuleConfig;
+
+        ProjectConfig {
+            modules: vec![
+                ModuleConfig::from_path("myorg.pack_a"),
+            ],
+            source_roots: [
+                "src/pack-a/src",
+                "src/pack-b/src",
+                "src/pack-c/src",
+                "src/pack-d/src",
+                "src/pack-e/src",
+                "src/pack-f/src",
+                "src/pack-g/src",
+            ]
+            .iter()
+            .map(PathBuf::from)
+            .collect(),
+            ..Default::default()
+        }
+    }
+
+    #[rstest]
+    fn check_internal_syntax_error_reports_as_error(
+        example_dir: PathBuf,
+        multi_package_config: ProjectConfig,
+    ) {
+        let project_root = example_dir.join("multi_package");
+        let result = check(&project_root, &multi_package_config, true, false).unwrap();
+
+        // Find the syntax error diagnostic
+        let syntax_error = result
+            .iter()
+            .find(|d| matches!(
+                d.details(),
+                DiagnosticDetails::Configuration(ConfigurationDiagnostic::SkippedFileSyntaxError { .. })
+            ))
+            .expect("Expected a syntax error diagnostic");
+
+        // The diagnostic should be an error, not a warning
+        assert!(matches!(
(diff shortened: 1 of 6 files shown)


src/commands/check/check_internal.rs
```diff
@@ -223,8 +223,8 @@ pub fn check(
                             ),
                         )]
                     }
-                    Err(DiagnosticError::ImportParse(_)) => {
-                        vec![Diagnostic::new_global_warning(
+                    Err(DiagnosticError::ImportParse(_)) | Err(DiagnosticError::PythonParse(_)) => {
+                        vec![Diagnostic::new_global_error(
```
DetachHead, line 227: i think the same fix is needed in `check_external.rs`:

```console
$ tach check-external
Configuration
⚠️  Warning: Skipped 'tach/asdf.py' due to a syntax error.
✅ All external dependencies validated!
```

ideally we should get rid of this duplicated code, see #853

DetachHead (commented):
thanks!

src/commands/check/check_internal.rs
DetachHead, line 227: a unit test for this would be good too

kimjune01:
Added a unit test — thanks for the suggestion.

src/commands/check/check_internal.rs
```diff
@@ -255,3 +255,66 @@ pub fn check(
...
+            ],
+            source_roots: [
+                "src/pack-a/src",
+                "src/pack-b/src",
+                "src/pack-c/src",
+                "src/pack-d/src",
+                "src/pack-e/src",
+                "src/pack-f/src",
+                "src/pack-g/src",
+            ]
```
DetachHead: prob want to put this syntax error fixture in a different source root, i think other tests are picking it up and failing as a result of the new module with a syntax error in it

kimjune01 (commented):
Done - moved to a dedicated syntax_error_fixture directory with its own tach.toml and module structure.

DetachHead:
to fix the type errors you can just add the test files to `ignore` and `exclude` https://github.com/tach-org/tach/blob/193eb5a03269516e4dfa4299e7d88fd497708adb/pyproject.toml#L112-L116

kimjune01:
Fixed. Added /** glob to the ignore/exclude patterns so pyright skips all files under syntax_error_fixture.

DetachHead (approved):
this looks good, just need to run `make fmt-rust` to fix the formatting issue.

in #933 i've updated the CI to not cancel the rest of the workflow if one check fails, which should make this less annoying

